PR #4, opened 2017-09-10
Add benchmarks of constant delay DDEs

devmotion:
As discussed in https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl/issues/3, I created notebooks with benchmarks of DDE algorithms. So far, I haven't benchmarked any DDEs with dependent delays since the models I deal with in my thesis only contain constant delays. However, DiffEqProblemLibrary already contains a set of interesting dependent delay problems.

I'm not sure whether my interpretation of the quantitative comparisons is always correct. Of course, I only compared a subset of methods with default options at different tolerances, so additional interesting comparisons might be missing.


ChrisRackauckas:
Benchmarks get refined over time. I think there are some very very interesting things here. 

Since the interpolants are used every step and they have extra steps, the cost of the higher order methods is proportionally higher, and so I'd think 5th order may constitute "high order" in this case, and maybe we should recommend `OwrenZen4`? Then recommend on the high end no more than `Vern6` or `Vern7`? Part of this probably is because we don't have Anderson acceleration though, which presumably would make larger steps faster and benefit higher order methods more.

I am surprised `OwrenZen5` takes that much longer though. I wonder if there's a bug or something like too big of a broadcast somewhere.

One comment I have for the future is that I like to keep a running "good method" through all of the tests. It helps establish a reference. So early on you show `OwrenZen4` doing well: keep it in the Verner method benchmarks so that way we can clearly see where the higher order methods then become a better choice.

devmotion:
> I am surprised OwrenZen5 takes that much longer though. I wonder if there's a bug or something like too big of a broadcast somewhere.

I was also surprised about this observation and thought there might be a bug somewhere because it is so much slower than all other RK methods.

> Since the interpolants are used every step and they have extra steps, the cost of the higher order methods is proportionally higher, and so I'd think 5th order may constitute "high order" in this case, and maybe we should recommend OwrenZen4? Then recommend on the high end no more than Vern6 or Vern7? Part of this probably is because we don't have Anderson acceleration though, which presumably would make larger steps faster and benefit higher order methods more.

`OwrenZen4` seems like a good default method, I guess. At low tolerances also `DP5` and `Tsit5` perform really well in both the Mackey and the Wheldon benchmarks. And on the high end, `Vern6` and `Vern7` seem like good choices. In these benchmarks I couldn't see any advantage in using `Vern8` or `Vern9`.

ChrisRackauckas:
> At low tolerances also DP5 and Tsit5 perform really well in both the Mackey and the Wheldon benchmarks. 

Yeah, that's the "free interpolant" at work. I wonder if there's a "simpler equation does better with OwrenZen" type of relation here, since they can cost more for their interpolation while more discontinuities will limit timesteps anyways.

I also see no advantage to the really high order methods, at least in "reasonable tolerance ranges".

PR #8, opened 2017-10-14
Update DDE benchmarks

devmotion:
I updated all DDE benchmarks on the current master branches of DelayDiffEq and OrdinaryDiffEq. Due to PRs https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/43 (higher order discontinuities) and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/209 (OwrenZen5 fixes) the current benchmarks are outdated. The new results are significantly different, especially for OwrenZen5 which was surprisingly slow before. It is unfortunate that OwrenZen5 now performs worse in the stiff QS benchmarks and also shows multiple oscillations in the qualitative benchmarks.


ChrisRackauckas:
Good stuff. The OwrenZen5 is now more in line with everything else which is good. Sad to see it's no longer "amazing" on that stiff QS benchmark though.

PR #11, opened 2018-01-22
Added notebooks for DiffEqBayes benchmarking

Vaibhavdixit02:
Notebooks of LotkaVolterra and Lorenz equations' parameter estimation using DiffEqBayes.


ChrisRackauckas:
1) To better show the data, you should show the true solution of the ODE run through the scattered data points. That then gives a sense of how much noise there has been added to the data.
2) The priors should be loosened up a bit. `Normal(mean,0.01)` is good for unit testing but for benchmarking they should be less strict so that it's given a chance to maybe mess up. Maybe perturb the mean a little bit, and make the variance higher (but keep them truncated).

Other than that, it's looking great!

ChrisRackauckas:
Plot using the plot recipe so that way the true continuous curve is shown instead of the linear interpolated version.

Add the uncertainty quantification plots to justify the tolerance choices.

Summarize it a little bit, describe what was faster and what performed better. 

Turing.jl needs longer chains on the LV model. You can see it hasn't converged. Is there a way to give it warmup samples?

Vaibhavdixit02:
Can you elaborate on the first point a little more?

ChrisRackauckas:
`plot(sol)` will use an interpolation formula to fill in the values between the solver's points in order to make the curves better match the true solution. 

http://docs.juliadiffeq.org/latest/basics/plot.html

Right now, you're plotting `sol.t` and `sol.u` directly which is linearly interpolating the two and thus doesn't accurately show what the solution looks like.

ChrisRackauckas:
There's no uncertainty analysis or Turing.jl for Lorenz

The priors are tight on Lorenz

There's no summaries.

ChrisRackauckas:
you'll want to show a slightly higher tolerance on Lorenz to show that low of a tolerance is necessary. Then I think it's done.

PR #16, opened 2018-02-12
DiffEqBayes Benchmarks for LV and Lorenz equation 

Vaibhavdixit02:


ChrisRackauckas:
Rename the README links too

PR #20, opened 2018-02-20
Update the Readme summary and typo in Lorenz notebook

Vaibhavdixit02:


README.md
```diff
@@ -146,3 +146,18 @@ over the set of tested equations and some specific examples may differ.
...
+
+### Parameter Estimation
+
+- Broadly two different approaches have been used, Bayesian Inference and Optimisation 
+  algorithms.
+- In general it seems that the optimisation algorithms perform more accurately but that can be 
+  attributed to the larger number of data points being used in the optimisation cases, Bayesian 
+  approach tends to be slower of the two and hence lesser data points are used, accuracy can 
+  increase if proper data is used.
+- Within the different available optimisation algorithms, BBO, GN_CRS2_LM, LD_SLSQP, LN_BOBYQA
```
ChrisRackauckas: Name which packages these are from, and mention the difference between global and local.

README.md
```diff
@@ -146,3 +146,18 @@ over the set of tested equations and some specific examples may differ.
...
+- Broadly two different approaches have been used, Bayesian Inference and Optimisation 
+  algorithms.
+- In general it seems that the optimisation algorithms perform more accurately but that can be 
+  attributed to the larger number of data points being used in the optimisation cases, Bayesian 
+  approach tends to be slower of the two and hence lesser data points are used, accuracy can 
+  increase if proper data is used.
+- Within the different available optimisation algorithms, BBO, GN_CRS2_LM, LD_SLSQP, LN_BOBYQA 
+  and LN_NELDERMEAD perform the best.
+- The different backends options available for Bayesian method offer some tradeoffs beteween
+  time, accuracy and control. It is observed that sufficiently high accuracy can be observed with
```
ChrisRackauckas: Do we have enough information to suggest heuristics yet?

Vaibhavdixit02: That is true, but at face value one can say something about the general trend though?

ChrisRackauckas: Oh no, I meant do we have enough information to suggest an in-category heuristic trend? Which of the 3 are generally faster, or more accurate?

Vaibhavdixit02: For faster we have the time from the `@time` macro and for accuracy I am just using an euclidean of sorts in my head I would say. I get that it isn't a concrete statement this way, would you prefer no mention of specific algorithms?

README.md
```diff
@@ -146,3 +146,21 @@ over the set of tested equations and some specific examples may differ.
...
+
+- Broadly two different approaches have been used, Bayesian Inference and Optimisation 
+  algorithms.
+- In general it seems that the optimisation algorithms perform more accurately but that can be 
+  attributed to the larger number of data points being used in the optimisation cases, Bayesian 
+  approach tends to be slower of the two and hence lesser data points are used, accuracy can 
+  increase if proper data is used.
+- Within the different available optimisation algorithms, BBO from the BlackBoxOptim package and GN_CRS2_LM, 
+  LD_SLSQP,LN_BOBYQA and LN_NELDERMEAD from the NLopt package perform the best.
+- The global versus local optimization make a huge difference in the timings. BBO tends to find the correct
```
ChrisRackauckas: The global ... => The choice of global ...

README.md
```diff
@@ -146,3 +146,21 @@ over the set of tested equations and some specific examples may differ.
...
+- Broadly two different approaches have been used, Bayesian Inference and Optimisation 
+  algorithms.
+- In general it seems that the optimisation algorithms perform more accurately but that can be 
+  attributed to the larger number of data points being used in the optimisation cases, Bayesian 
+  approach tends to be slower of the two and hence lesser data points are used, accuracy can 
+  increase if proper data is used.
+- Within the different available optimisation algorithms, BBO from the BlackBoxOptim package and GN_CRS2_LM, 
+  LD_SLSQP,LN_BOBYQA and LN_NELDERMEAD from the NLopt package perform the best.
+- The global versus local optimization make a huge difference in the timings. BBO tends to find the correct  
+  solution for a global optimization setup. For local optimization, most methods in NLopt, like :LN_BOBYQA, solve the
```
ChrisRackauckas: You should probably mention `GN_CRS2_LM` is global, while the others are local.

README.md
```diff
@@ -146,3 +146,21 @@ over the set of tested equations and some specific examples may differ.
...
+  algorithms.
+- In general it seems that the optimisation algorithms perform more accurately but that can be 
+  attributed to the larger number of data points being used in the optimisation cases, Bayesian 
+  approach tends to be slower of the two and hence lesser data points are used, accuracy can 
+  increase if proper data is used.
+- Within the different available optimisation algorithms, BBO from the BlackBoxOptim package and GN_CRS2_LM, 
+  LD_SLSQP,LN_BOBYQA and LN_NELDERMEAD from the NLopt package perform the best.
+- The global versus local optimization make a huge difference in the timings. BBO tends to find the correct  
+  solution for a global optimization setup. For local optimization, most methods in NLopt, like :LN_BOBYQA, solve the  
+  problem very fast.
```
ChrisRackauckas: but require a good initial condition

README.md
ChrisRackauckas: I would prefer a mention of what's fast and what's accurate if you think they are already starting to differentiate from each other.

Vaibhavdixit02:
@ChrisRackauckas does it look alright now? Do you want to change the comparison in accuracy?

PR #26, opened 2018-05-08
jump benchmarks

isaacsas:
Some SSA benchmark notebooks.

I've left in the BenchmarkTools timing for reference, but I think the `@elapsed` results are more consistent. With that said, in all the benchmarks I've run something funny happens with the Marchetti gene expression example, where the direct method gives extremely slow results the first time benchmarking is run (the first 10 or so calls to `solve`). Subsequently re-running the benchmark seems to fix everything. This happens with both BenchmarkTools and using `@elapsed`.


ChrisRackauckas:
These look great. I'll merge so we can get the nbviewer links and then add them to the README.

BTW, is there any way that we can call out to some other libraries, maybe through RCall or PyCall, to evaluate the timing difference?

isaacsas:
Most of the popular SSA software (BioNetgen, StochSS, Copasi) is provided as compiled binaries or buildable binaries, without much in the way of nice API interface (they assume users want to load some type of gui or just supply an SBML input file generated from a gui, and then provide executables to run the simulations and dump trajectories to a file). They can probably be scripted from the command-line once appropriate input files are setup. 

The one exception is pSSALib, which implements the partial propensity methods. They seem to also provide a C++ based API. I was actually thinking it might make sense to wrap their library rather than implementing my own partial-propensity solvers (they are pretty involved, and also restrictive in the types of rate expressions they allow, so less general than most of the methods I plan to look at). I have no idea how easy that would be to do though given that it is C++...

There are a few Python-based SSA codes, I think the only one I know of that has good performance calls the StochKit executable (which StochSS also uses), it is called Gillespy.py. I don't know about R.

isaacsas:
The Mendes paper I've referenced before includes it in their benchmarks, and shows it is (about?) the worst performing SSA library of all those they tested. So at least some people know this. I agree though that at some point comparing with it and some of the faster C/C++ codes would be good. 

I guess I'm more interested in knowing how much worse we are doing then the fastest codes, and trying to close the gap. (Since ultimately I want to build spatial SSAs in Julia and actually use them in my research. I'd like to spend much less of my time writing C++ code...) 

For example, I'm pretty sure we are getting much worse performance of the NRM due to the priority queue implementation than C/C++ codes get. I'd like to understand what is slowing down the queue updating relative to C/C++. In the end, they're all just swapping entries in a couple vectors to make updates to the datastructure...

ChrisRackauckas:
>I guess I'm more interested in knowing how much worse we are doing then the fastest codes, and trying to close the gap. (Since ultimately I want to build spatial SSAs in Julia and actually use them in my research. I'd like to spend much less of my time writing C++ code...)

Indeed. I agree that measuring against C++ code is more useful for development. I think there's two perspectives here. The development perspective is trying to understand where the field is so we can directly improve by changing our strategies to match the better parts of each. This is what benchmarking against the C++ code is for. The other perspective is growing the organization and showing potential users where we place in the grand scheme of things. This is where having published and easily verifiable notebooks are a useful.

PR #32, opened 2018-11-18
RFC: Restructure benchmarks to use Weave instead of notebooks

ChrisRackauckas:
What runs:

- [x] Adaptive SDE
- [x] Dynamical ODE
- [x] Jumps
- [x] Nonstiff DDE
- [x] Non-stiff ODEs
- [x] Non-stiff SDE
- [x] Parameter estimation
- [x] Stiff DDE
- [ ] Stiff ODE
- [x] Stiff SDE

What's been ran on the PuMaS computer:

- [x] Adaptive SDE
- [x] Dynamical ODE
- [ ] Jumps
- [x] Nonstiff DDE
- [x] Non-stiff ODEs
- [x] Non-stiff SDE
- [x] Parameter estimation
- [x] Stiff DDE
- [ ] Stiff ODE
- [x] Stiff SDE


isaacsas:
Is the idea here that Weave will play nicer with git? I like being able to view the notebooks online, with Weave will we be able to store snapshots of completed benchmark studies that can be viewed online?

(I do agree that notebooks don't work super well with git, so maybe Weave + some other method to store completed benchmark results would be good.)

ChrisRackauckas:
This generates a script, html, and a pdf. The script is a .jl file with all of the commands and the text included as comments. The html is a good web display of the results:

http://htmlpreview.github.io/?https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl/blob/3d633c069db8338467172acfe191093ed9621d0b/html/NonStiffODE/linear_wpd.html

and the pdf is a shareable version of the results:

https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl/blob/3d633c069db8338467172acfe191093ed9621d0b/pdf/NonStiffODE/linear_wpd.pdf

So these are all still viewable online. The .jmd plays much nicer with Git though, it plays well with Juno (and probably more support coming elsewhere. In Juno it just acts as a runnable .jl file where you can execute lines and blocks). And this also makes it easier to do programmable parts like the footer which will include information about the computer, how to run the script, and the current package state.

ChrisRackauckas:
This is what the completed example looks like with the footer: https://htmlpreview.github.io/?https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl/blob/9f2c1a96db339251e00eaf825ae60c9195070a66/html/NonStiffODE/linear_wpd.html

YingboMa:
Switching to `Weave.jl` has no objections from me. Just one very minor detail, I suggest using `plot(..., dpi=200)` to achieve higher quality plots.

YingboMa:
But in the generated HTML file, it still uses the PNG format.

ChrisRackauckas:
Ahh I didn't catch that. Yes, we should change that.

ChrisRackauckas:
Conversion was done via:

```julia
for folder in readdir(joinpath(repo_directory,"benchmarks"))
  folder == "test.jmd" && continue
  for file in readdir(joinpath(repo_directory,"benchmarks",folder))
    file[end-4:end] != "ipynb" && continue
    println("Converting $(joinpath(folder,file)))")
    convert_doc(joinpath(repo_directory,"benchmarks",folder,file),
                joinpath(repo_directory,"benchmarks",folder,file[1:end-5]*"jmd"))
  end
end
```

ChrisRackauckas:
SDEs should be able to run too. Doing some quick checks.

ChrisRackauckas:
For non-stiff SDE, it's just the weak wpd

ChrisRackauckas:
- [ ] AdaptiveEfficiencyTests needs to be re-ran
- [ ] ~2x speed-downs are seen in Oval2LongRun and Oval2LongTimes. @YingboMa we need to `@..` StochasticDiffEq.jl since it's using broadcast everywhere. Note that the implicit methods are more stable.
- [ ] Oval2Timings needs to be re-ran
- [ ] qmaxDetermination needs to be re-ran

Henon-Heiles was fine

- [ ] Quadrupole_boson_Hamiltonian_energy_conservation_benchmark needs to be re-ran
- [ ] single_pendulums needs to precompile

- [ ] nonstiff DDEs improved. Need to be re-ran for precompile.

- [ ] Pleiades_wpd needs to be re-ran with the error message removed.
- [ ] ThreeBody_wpd needs to be re-ran with the error message removed.

- [ ] Weak SDE needs to be re-ran
- [ ] BasicSDEWorkPrecision shows a regression, possibly needs `@..` everywhere.
- [ ] LotkaVolterraSDE RKMil is wonky. Shows a regression.

- [ ] DiffEqBayesLorenz and DiffEqBayesLotkaVolterra need to be re-ran

QuroumSensing improved. MethodOfSteps(Rodas5) is now more sensitive to tolerances, diverging at 1e-3 on QuorumSensing. Tolerances were changed to reflect this.

- [ ] Filament needs to be fixed

The stiff ODE solver benchmarks are fine timing-wise. In fact, everything improved, even the Sundials and LSODA wrappers (could be due to the new benchmarking computer to some extent). However, we need to check the effect of BLAS threading and W-matrices.

- [ ] There are massive performance regressions in StochasticHeat.

QuadraticStiffness is looking great!

PR #49, opened 2019-07-31
add extrapolation methods

saurabhkgp21:


benchmarks/NonStiffODE/linear_wpd.jmd
```diff
@@ -213,6 +213,19 @@ wp = WorkPrecisionSet(prob,abstols,reltols,setups;names=solnames,
 plot(wp)
 ```
 
+## Comparison with Extrapolation methods
+
+Now let's test Tsit5 against parallel extrapolation methods:
+```julia
+setups = [Dict(:alg=>Tsit5())
+          Dict(:alg=>AitkenNeville(min_order=1, max_order=9, init_order=9, threading=true))
+          Dict(:alg=>ExtrapolationMidpointDeuflhard(min_order=1, max_order=9, init_order=9, threading=true))]
```
ChrisRackauckas: Why is Hairer-Wanner missing? Isn't that the one that usually does best?

saurabhkgp21: Bug in HairerWanner isn't yet solved. Should I add HairerWanner now or add it after solving that bug?

ChrisRackauckas: wasn't this working before?

saurabhkgp21: HairerWanner will work for this problem, but its not guaranteed to solve any problem we will throw to it. Adding HairerWanner here will work for now, I think. So should I add HairerWanner in benchmarks?

PR #54, opened 2019-10-01
Update jump benchmarks

isaacsas:
Starting a PR for updating jump benchmarks.


isaacsas:
@ChrisRackauckas I think I've now got the four older jump benchmarks updated. So this should hopefully be ok to merge (I generated the html for each on my computer without issues.) As we discussed I've only included the Weave file here, but can add html and scripts if that is not autogenerated.

I'll add the BCR one in a separate PR when I get a chance to convert it to Weave.

PR #81, opened 2019-11-25
Save all figures as SVG

asinghvi17:
Improves resolution and text


ChrisRackauckas:
Thanks!

ChrisRackauckas:
Can we somehow automatically do this to all?

asinghvi17:
I dont see why not...you just locate the end of the header and insert this text before it

ChrisRackauckas:
yeah, but I think we can template these somehow?

asinghvi17:
We can also pass this option to the call to `weave`.  I don't think I can run the examples locally though.

PR #85, opened 2019-12-12
Update BCR benchmark to use correct path

isaacsas:
I'm not sure if the auto-generated Jacobian test can actually go through when using autodiff, so this may need an update to force using finite difference Jacobians.


isaacsas:
Also, I haven't tested @ChrisRackauckas' benchmarking code, but I assume that works ok.

isaacsas:
Accidentally did this on master. If you prefer it as a new branch let me know and I'll resubmit.

PR #101, opened 2020-05-09
Use `truncated` instead of `Truncated`

devmotion:
Just noticed that DiffEqBenchmarks uses the deprecated `Truncated` instead of `truncated` which (potentially) creates optimized truncated distributions and uses `Truncated` as fallback. I remember some simple models in Turing being quite slow due to deprecation warnings, so probably better to use `truncated` in the benchmarks.

I'm not sure what's the correct procedure for rerunning and regenerating the different output formats?


ChrisRackauckas:
The procedure to run it is documented in https://github.com/SciML/DiffEqBenchmarks.jl#contributing .  It might be good to re-weave all 3 of those. @Vaibhavdixit02 and I have been discussing the Lorenz one, because Stan fails with those priors by just never finishing, so we are trying to find a good way to document that behavior, like cut it early or something. If you have any ideas go for it.

PR #106, opened 2020-06-04
Add NBodySimulator benchmarks

SebastianM-C:
- [x] Low level benchmarks
   - [x] Gravitational potential
   - [x] Coulomb potential
   - [x] Magnetic dipole
   - [x] Lenard Jones potential
   - [x] Harmonic bond potential
   - [x] Valence angle potential
- [ ] High level benchmarks
   - [x] Liquid argon simulation
   - [ ] Water simulation

The low level part, statistical benchmarks should test the acceleration functions in order to help catch possible performance regressions and provide a baseline for future optimizations.
For the high level simulation I think work-precision-like diagrams would be of interest, focusing on
 - time vs energy error
 - number of steps vs energy error
for various integrators.


SebastianM-C:
I marked this as ready by mistake.
It is almost ready, but I was wondering what integrators are most relevant for the list. Currently the first benchmarks runs without problems, but takes a few hours. The second hangs for some yet unknown reason at `τ=3.1622776601683796e-14`.

ChrisRackauckas:
I'd try a smattering of symplectic methods, DPRKNs, and then throw Tsit5, Vern7, and Vern9 in there for good measure

SebastianM-C:
I addressed the previously mentioned problems and calibrated the timestep by computing a cost factor that should take into account the cost per timestep between between different integrators.
What I noticed now is that there are some (significant I'd say) differences between the weaved files, namely the energy error doesn't look the same between the runs. In the html file we have
image
and
image

while in the markdown file
image
and
image

and in the pdf
image
and
image

I'm not sure how to explain these differences. The weaving process happens on the same machine in all cases and with the same code.

SebastianM-C:
The initial conditions are not exactly the same, since velocities are random

https://github.com/SciML/NBodySimulator.jl/blob/172b046d8f3cfd0737df294fb54481b0bc95e754/src/bodies.jl#L35

SebastianM-C:
The variation is due to the computation of the cost (which depends on runtime) at each run. The best way to solve this is to compute the cost once and then hardcode them.

SebastianM-C:
With fixed time-step cost, here are some new results
image

`McAte2` seems to consistently perform better than other solvers. I also noticed that it is the single solver with a lower cost than `VelocityVerlet` and I was wondering if this is relative to it's performance relative to `VelocityVerlet`.

In the energy error history I noticed that high order solvers then to have a higher energy error (and slightly lower runtimes) and I'm not sure if this indicates problems with the cost tuning.
image

With the adaptive solvers I also tried to apply the same line of reasoning in regards to cost tuning, but I still get the same behavior of constant runtime.
image

Comparing to the symplectic solvers it seems that `DPRKN6` and `DPRKN8` are quite competitive.

ChrisRackauckas:
the errors on the fixed time step ones still look fairly random. I'm surprised

SebastianM-C:
I think we can leave the water simulation for later. This should hopefully produce better results. I'll also start a local run to check again if everything is ok. There are also some long time benchmarks included, but commented out (with `t=100`). From what I understand form the literature, we'd need something with ~10^8 Verlet timesteps for talking about long term, so maybe a longer runtime is needed. I did not test that due to how long it would take. If you want, I can enable them, but the runtimes could well be over a week or more (since everything is run 3 times).

ChrisRackauckas:
It would be fine to get that setup as a separate file that we don't run often.

PR #108, opened 2020-06-15
Rebuild benchmarks

christopher-dG:
This is dependent on https://github.com/JuliaGPU/gitlab-ci/pull/22

Basically the same thing as the tutorials.

TODO:
- Figure out which runner tag should be used for benchmark jobs
- Change a path from `cdg/rebuilds` to `master` in `.gitlab-ci.yml` when the JuliaGPU/gitlab-ci PR is merged

Here's a commit it made: https://github.com/SciML/DiffEqBenchmarks.jl/commit/c211b96ea685bc122e865eb45458f0d63fd4af87

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,7 +1,7 @@
 name: TagBot
 on:
   schedule:
-    - cron: 0 * * * *
+    - cron: 0 0 * * *
 jobs:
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
>Figure out which runner tag should be used for benchmark jobs

nvidia-benchmark

PR #109, opened 2020-06-26
Move to separate projects

christopher-dG:
I did this manually by greping for `using` and `import` in each folder, I may have missed some.

Also included the fix for broken PDF generation that we did in Tutorials.


ChrisRackauckas:
That's more accurate than I would've done it. I would've just done it by hand and 🤞 . So I'll just merged, and we can start regenerating the benchmarks. They needed to get updated anyways. We'll see what happens when things stall. Let's do these one-by-one though since some will take a really long time. Nonstiff ODEs first.

Issue #116, opened 2020-07-04
Build log


ChrisRackauckas:
!rebuild NonStiffODE

christopher-dG:
!rebuild NonStiffODE

it's gotta be an issue comment

ChrisRackauckas:
!rebuild StiffODE

ChrisRackauckas:
!rebuild Jumps

ChrisRackauckas:
!rebuild NonStiffDDE

ChrisRackauckas:
!rebuild NonStiffSDE

ChrisRackauckas:
!rebuild ParameterEstimation

ChrisRackauckas:
!rebuild StiffDDE

ChrisRackauckas:
!rebuild StiffSDE

ChrisRackauckas:
🍿

ChrisRackauckas:
!rebuild AdaptiveSDE

ChrisRackauckas:
!rebuild DynamicalODE

ChrisRackauckas:
!rebuild NonStiffDDE

ChrisRackauckas:
!rebuild StiffDDE

ChrisRackauckas:
!rebuild DynamicalODE

ChrisRackauckas:
!rebuild NBodySimulator

ChrisRackauckas:
!rebuild Bio

PR #193, opened 2021-04-01
[WIP] This adds the buildkite setup

ven-k:
- Checks if the folder in `benchmarks` has `.jmd` file
- Creates a pipeline for each `.jmd` file
- If there is any change in the rest of the repo runs the tests
- Uploads pdf, html... as artifacts


ChrisRackauckas:
Changes:

- Make trigger like https://github.com/SciML/SciMLBenchmarks.jl/issues/116
- Setup with the AMDCI3 tags
- Copy same thing over to SciMLTutorials
- Allow marking the GPU benchmarks to use gpuci1

ChrisRackauckas:
@staticfloat mentioned we should go away from text triggers. What do you recommend? Our biggest issue here is that these benchmarks should hardly ever be updated, but should be run often.

ven-k:
We can have a workflow dispatch ([a demo at 1 min](https://dev.to/github/github-actions-manual-triggers-with-workflowdispatch-3di1) - this can be modified to accept (folder, file) options)

That's one of the 2 ways to manually trigger the workflow through Github actions. (The other one is `repository_dispatch` for integration if there are GUIs). This workflow can have buildkite pipeline (maybe[ this](https://github.com/marketplace/actions/trigger-buildkite-pipeline#:~:text=Usage,the%20settings%20for%20the%20build) plugin is useful there) 

Buildkite's options to trigger didn't seem this flexible (considering "no-text" triggering)

.github/workflows/update.jl
```diff
@@ -0,0 +1,24 @@
...
+    cd(model_dir)
+    Pkg.activate(".")
+    Pkg.update()
+    Pkg.activate()
+    manpath = (joinpath(benchpath, "Manifest.toml"))
+    if length(readlines(`git status . --porcelain`)) > 1
+        run(`git checkout -B $dir`)
+        run(`$git add -A . :!manpath`)
+        run(`$git commit -m "Updated $dir"`)
+        run(`$git push --set-upstream origin $dir`)
```
ven-k, line 21: Would `git request-pull $dir origin master` create a PR?

ven-k:
@staticfloat as "Request a reviewer" isn't showing up, I'm tagging you for a review.

.buildkite/rebuild_models.yml
```diff
@@ -0,0 +1,43 @@
...
+steps:
+  - label: ":runner: Dynamically launch Pipelines"
+    plugins:
+      - staticfloat/forerunner:
+          # This will create one job per file
+          watch:
+            - path: benchmarks/.*\.jmd
+          path_processor: per-file
+          target: .buildkite/coalescer/rebuild_models
+          target_type: template
```
staticfloat, line 13: Can you explain to me what you're trying to do here?

ven-k: I want to watch the `.jmd` files. If there is any modification then run the benchmarks for just that file.

The `.buildkite/coalescer/rebuild_models` currently creates a `MODELS` which has "folder_A/modified-file.jmd" as entries. Currently, the script only does an `echo` (should I do something to pass this) and it is supposed to be used by `rebuild_models.jl` to `weave_file(folder, file)`.

Earlier, something like `!rebuild folder_A` would make it check for the `modified.jmd` and launch a pipeline for each of them.

.buildkite/commands/rebuild_models.jl
```diff
@@ -0,0 +1,15 @@
...
+model = ARGS[1]
+folder, file = split(model, "/")
+
+project_abspath = joinpath(repo_dir, "library", model)
+if !isfile(project_abspath) || !occursin(file, ".jmd")
+    error("Invalid file $file")
+end
+
+@info("Rebuilding $folder/$file")
+weave_folder(folder, file)
```
ven-k, line 15: This has to be `weave_file(folder, file)`. Will fix it in a subsequent commit

ChrisRackauckas:
Resolve conflicts and let's merge. Let's just see what happens.

PR #199, opened 2021-04-14
This adds files to weekly update all the directories.

ven-k:
- If modifications are found, it creates a new branch with dirname

- As it's a cron job, it might be necessary to merge to test if the authentication in `update.yml` works
- Currently, cron is set to run once every hour (and will be changed once fully tested)

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,9 @@
+steps:
+  - label: ":runner: Dynamically launch Pipelines"
+    plugins:
+      - staticfloat/forerunner#0c78044b02ced95d8b9ca6fda7883fba45862fd2:
+          # This will create one job overall, throwing all path information away
+          watch:
+            - path: src/*.*.jl
+            - path: .*\\.toml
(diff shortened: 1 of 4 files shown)


ven-k:
@ChrisRackauckas the recent PRs (#201, #200) have failed as there is no `pipeline.yml` for the Buildkite webhook to work with. So this PR, along with 
```
- update.yml file invokes update.jl on workflow_dispatch or on Sat 6 am UTC
- Peek in each dir in benchmarks
- If there are any changes push it to a "dir" branch (create/ update if it already exists)
- Create a PR (or update if one already exists)
```
, now, adds a `test_sciml.yml` to test the overall changes done for the repo itself on the Buildkite. 

(The Buildkite setup will be eventually updated when #193 is merged i.e with a commit changing how artifacts-path is passed- if nec)

ChrisRackauckas:
Is there a list of folders to exclude with this?

ven-k:
No. It updates any folder with changes in `benchmarks`

PR #213, opened 2021-04-24
[WIP] A separate watcher for each file

ven-k:
- 2 test commits will follow to ensure it works

### DIFF
--- a/.buildkite/commands/rebuild_models.jl
+++ b/.buildkite/commands/rebuild_models.jl
@@ -2,16 +2,6 @@
 
 using Pkg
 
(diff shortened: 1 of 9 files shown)


ven-k:
@ChrisRackauckas, it watches and triggers pipeline for both files now.

ChrisRackauckas:
Seems to be a real test failure?

staticfloat:
So I think I fixed the error that Venkatesh was having; but there seems to be an actual test error here when running the package tests themselves.

ChrisRackauckas:
https://github.com/JunoLab/Weave.jl/issues/422

PR #220, opened 2021-05-03
Fix activation

staticfloat:


src/SciMLBenchmarks.jl
```diff
@@ -8,8 +8,8 @@ function weave_file(folder,file,build_list=(:script,:html,:pdf,:github,:notebook
   target = joinpath(folder, file)
   @info("Weaving $(target)")
   
-  if isfile(joinpath(target, "Project.toml"))
-    @info("Instantiating...")
+  if isfile(joinpath(folder, "Project.toml"))
+    @info("Instantiating", folder)
```
ChrisRackauckas, line 12: I think this is the style where now SciMLBenchmarks won't be defined?

ChrisRackauckas: Yup the output has that.

ChrisRackauckas:
More Sundials fixes are in https://github.com/SciML/Sundials.jl/pull/305

PR #226, opened 2021-05-09
bump nonstiff ODE packages for FastBroadcast.jl

ChrisRackauckas:


ChrisRackauckas:
@chriselrod @YingboMa it looks like that fixed the performance regression.

Before:

before

v1.6:

v16

After adding FastBroadcast.jl:

after

🎉

chriselrod:
You edited to remove the plots; are they accurate?

chriselrod:
Ha, yes, before and v1.6

PR #229, opened 2021-05-09
update DynamicalODE benchmarks

ChrisRackauckas:


ChrisRackauckas:
@SebastianM-C could you help in updating this one?

SebastianM-C:
Sure. I was wondering if we should try to take into account the time step cost like in the N-body benchmarks.

ChrisRackauckas:
total time is fine, if it's compared with error.

PR #232, opened 2021-05-09
update NBodySimulator benchmarks

ChrisRackauckas:


ChrisRackauckas:
@SebastianM-C could you look into why this one doesn't plot anymore?

PR #233, opened 2021-05-09
update NonStiffDDE benchmarks

ChrisRackauckas:


ChrisRackauckas:
@devmotion @YingboMa @chriselrod this benchmark shows a regression that would be good to identify, since most benchmarks seem to have improved. Seems DelayDiffEq-specific.

PR #242, opened 2021-05-12
Bump parameter estimation benchmarks

ChrisRackauckas:


ChrisRackauckas:
@staticfloat how do I retrigger a benchmark for a folder?

staticfloat:
If you modify a `.toml` file, it should rebuild everything within the folder.

PR #230, opened 2021-05-09
update Jumps benchmarks

ChrisRackauckas:

PR #275, opened 2021-05-15
update Jump benchmarks to use Catalyst.jl

ChrisRackauckas:


Gregliest:
So it looks like there's a previous bug, not related to the upgrade to Catalyst. The [errors](https://julialang-buildkite-artifacts.s3.amazonaws.com/scimlbenchmarks/dbb53a9455a7d6722cdf1d333625c600b09a5581a9c1c4ad7ef06364a4afbf97/html/Jumps/Mendes_multistate_example.html) in the Mendes Multistate benchmark look like they were present in [this PR ](https://github.com/SciML/SciMLBenchmarks.jl/pull/230) on 5/15. 

The artifacts for [this PR](https://github.com/SciML/SciMLBenchmarks.jl/pull/218) show that it was working as of 5/2.

Unfortunately, a lot happened between the two PRs. Since it doesn't look like the code in the source file has changed in that time, I'm assuming it's some sort of dependency issue. I'm in the process of trying to figure out what the code is trying to do, and what may have changed in the dependencies, but I'm obviously not that familiar with the ecosystem yet. Do you have any quick thoughts on what might be the problem? Otherwise I'll keep diving into `ModelingToolkit.ReactionSystem` and `JumpProblemNetwork`.

ChrisRackauckas:
>The artifacts for this PR show that it was working as of 5/2.

That's the old pre-Catalyst DiffEqBio form. I think it should be `nameof.(states(rn))`, but then the symbol names may need to change. The better way to do this would be to completely remove that handling and instead do:

```julia
@variables t S7(t) S8(t) S9(t)
```

and then further down use

```julia
solv[:,i] = sum(sol[S7], dims=1)
```

since in modern symbolic ModelingToolkit you can just index by name.

PR #286, opened 2021-06-11
Update 2 of the Jumps benchmarks to use Catalyst

Gregliest:
Continuation of https://github.com/SciML/SciMLBenchmarks.jl/pull/275. 

This PR fixes NegFeedback_GeneExpr.jmd and NegFeedback_GeneExpr_Marchetti.jmd. I'm having some trouble getting the other two bencharks to work, so I thought I'd submit this in the mean time. The outputs are slightly different but similar to the original outputs, which I'm assuming is due to randomness. However, I still don't have a great grasp of what's being tested, so someone should double check that the values are correct. 

I also couldn't 100% verify what `save_positions(false, false)` was doing. Is there some behavior I need to add back for that? 

For the other two I'm struggling with some weird errors and behavior, which I'll post separately if I can't figure out.

### DIFF
--- a/benchmarks/Jumps/Diffusion_CTRW.jmd
+++ b/benchmarks/Jumps/Diffusion_CTRW.jmd
@@ -4,7 +4,7 @@ author: Samuel Isaacson, Chris Rackauckas
 ---
 
 ```julia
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
Cool, I cut the artifacts down to the two fixed ones to merge this. That's definitely a big improvement, thanks!

PR #289, opened 2021-06-13
Fix Mendes Benchmark

Gregliest:
Fixes the Mendes benchmark and updates ModelingToolkit to allow the use of `save_positions` kwarg. 

A couple of notes:
- I explicitly added ModelingToolkit as a dependency to get the correct version. Not sure if that will be an issue in the future or if there was a better way to force that.
- I couldn't get the symbol based indexing to work for the Mendes problem, as it appears that the problem from the problem library doesn't use them. I didn't spend too much time looking into it though.


ChrisRackauckas:
This looks good, I think we need to think about a few of the symbolic indexing issues a bit more. It should just work if you do:

```julia
@variables t x(t)
sol[x]
```

but 🤷 we need some way to better reference the names of the reactions and lower them nicely.

Gregliest:
Maybe the variables should be a property of something, like either the `System` or the `Problem`? In this case in particular, since we're getting the problem from the problem library, you shouldn't be able to declare an arbitrary variable like `@variables t foo(t)` and expect sol[foo] to work. 

In this case, it appears that the `sol` doesn't have any indexing symbols, as `sol.prob.f.syms` is empty. It's kind of strange, because `plot(sol)` has a legend with `u1(t), u2(t), ...`, but doing `@variables t u1(t)` then `sol[u1]` didn't work for me.

PR #298, opened 2021-08-12
[WIP] Sanft

Vilin97:
Added a benchmarking script for spatial SSAs using the model from Sanft, 2015.


ChrisRackauckas:
What would it take for this to get finished?

Vilin97:
Benchmarking flattened SSAs seems to result in segfaults, unclear why. Also, the pdf does not seem to get built either. I don't know how to resolve either of those two issues.
@isaacsas , do you have any insight? We looked into this at the end of August but couldn't figure out back then.

isaacsas:
I think pdfs aren't getting built anymore, so that issue should hopefully go away.

I need to run this locally and see if I can reproduce the crashes on the flattened systems, hopefully soon.

ChrisRackauckas:
Rebase so it uses the latest build scripts.

isaacsas:
Updating this in https://github.com/SciML/SciMLBenchmarks.jl/pull/408 so can close here.

PR #308, opened 2021-09-20
update BCR and its manifest

isaacsas:


ChrisRackauckas (approved):

ChrisRackauckas:
I'm curious how this turns out.

isaacsas:
Me too! I also want to start keeping an eye on the ODE rhs timing to help avoid issues like that recent one (where we worried about some kind of fictitious `build_function` regression).

isaacsas:
Closing and reopening doesn't restart buildkite?

isaacsas:
Oh, should I bump the manifest again for stuff that was merged in Symbolics/SymbolicUtils yesterday?

ChrisRackauckas:
yeah probably best to bump

isaacsas:
At least the html finished this time https://julialang-buildkite-artifacts.s3.amazonaws.com/scimlbenchmarks/3204bd4da0e46ad91a1f57e6775dae5b917e2621780719cff3cf9855374cb0b5/html/Bio/BCR.html

Interesting there is a bit of allocation in calling the ODE rhs function according to btime.

ChrisRackauckas:
Yeah, let's get an MWE of this and investigate it.

isaacsas:
Here is a MWE 
```julia
rn = @reaction_network rn begin
              (k1,k2), A <--> B
              end k1 k2
osys = convert(ODESystem,rn)
oprob = ODEProblem(osys,states(osys) .=> [1.0,1.0],(0.0,10.0), parameters(osys) .=> [1.0,1.0])
du = zeros(2); u = [1.0,1.0], p=[1.0,1.0];
```
timing gives
```julia
julia> @btime oprob.f($du,$u,$p,0.0)
  131.663 ns (3 allocations: 512 bytes)
```
Or does `oprob.f` need to be interpolated? Doing that gives no allocations but times seem too small (like 2 ns).

isaacsas:
This goes away if one directly generates the function
```julia
f = generate_function(osys, expression=Val{false})[2]
julia> @btime f($du,$u,$p,0.0)
  18.054 ns (0 allocations: 0 bytes)
```
The generated code is
```julia
:(function (ˍ₋out, ˍ₋arg1, ˍ₋arg2, t)
      let var"A(t)" = @inbounds(ˍ₋arg1[1]), var"B(t)" = @inbounds(ˍ₋arg1[2]), k1 = @inbounds(ˍ₋arg2[1]), k2 = @inbounds(ˍ₋arg2[2])
          @inbounds begin
                  ˍ₋out[1] = (+)((*)(k2, var"B(t)"), (*)(-1, k1, var"A(t)"))
                  ˍ₋out[2] = (+)((*)(k1, var"A(t)"), (*)(-1, k2, var"B(t)"))
                  nothing
              end
      end
  end)
```

ChrisRackauckas:
Oh that seems like it's just a global scope `getproperty` thing. Looks fine.

isaacsas:
Cool then this should be good. It only crashed as it was running the pdf build and went over the time limit. The html build finished fine.

PR #322, opened 2021-11-23
Move CI to `juliaecosystem`

staticfloat:
This setup mostly mirrors that of `SciMLTutorials`


ChrisRackauckas:
Both this repo and the other have a weird test failure since some file path moved. I can't quite figure it out, but if you have an idea that would be helpful. For now I'm just merging when I see that failure get that far though.

PR #327, opened 2021-11-27
Add NeuralPDE Paper Benchmarks

ChrisRackauckas:
This adds the benchmarks from https://arxiv.org/abs/2107.09443 to the repository.

These should probably get setup to use GPUs but for now I think this should all train fine on CPU.


ChrisRackauckas:
I'm just going to go ahead and merge and see what ends up building. @KirillZubov and @zoemcc follow up on these. We can use these as baselines in the future.

PR #355, opened 2021-11-30
Fix Allen Cahn pinn

KirillZubov:


ChrisRackauckas:
Won't all ErrorsVsTime benchmarks need the same fix?

https://benchmarks.sciml.ai/html/PINNErrorsVsTime/diffusion_et.html

etc.

KirillZubov:
@ChrisRackauckas  yes, all, but I began from one :)

ChrisRackauckas:
okay cool. Let's see how this builds.

PR #356, opened 2021-12-10
Update and fix parameter estimation benchmarks

ChrisRackauckas:
@devmotion do you know why Turing is failing at precompilation?


devmotion:
Libtask_jll is not (and probably will never be) compatible with Julia 1.7, see eg https://github.com/TuringLang/Turing.jl/issues/1738#issuecomment-988421993.

ChrisRackauckas:
@staticfloat I updated the benchmarks to v1.7 but they didn't spawn... did I miss a spot? Or is this something to do with the configuration of the agents?

PR #371, opened 2021-12-14
Fix a bunch of PINN benchmarks

ChrisRackauckas:


benchmarks/PINNErrorsVsTime/allen_cahn_et.jmd
```diff
@@ -183,7 +182,7 @@ strategies = [NeuralPDE.QuadratureTraining(quadrature_alg = CubaCuhre(), reltol
               NeuralPDE.QuadratureTraining(quadrature_alg = CubatureJLp(), reltol = 1, abstol = 1e-4, maxiters = 100),
               NeuralPDE.GridTraining(0.09),
               NeuralPDE.StochasticTraining(100),
-              NeuralPDE.QuasiRandomTraining(100; sampling_alg = UniformSample(), minibatch = 100)]
+              NeuralPDE.QuasiRandomTraining(100; minibatch = 100)]
```
KirillZubov, line 185: a little lower 
#LBFGS -> LBFGS
```julia
minimizers = [ADAM(0.005)],
              #BFGS()]
              LBFGS()]
minimizers_short_name = ["ADAM"]
                        "LBFGS"]
                        # "BFGS"]
```

ChrisRackauckas:
@KirillZubov how long do you expect this to run? It looks like one got stuck for 16 hours.

KirillZubov:
I propose to first run with a small number of iterations to check that everything works, and then return to the full number of iteration

KirillZubov:
maxIters = [(**1,1,1,1,1,**20000,20000),(300,300,300,300,300,300,300)]  
it is not clear why 1 iteration. Here, or all strategies have the same number of iteration for full convergence or just need to remove a part of strategies

ChrisRackauckas:
Can you modify those?

KirillZubov:
ok,  I'll do it

ChrisRackauckas:
Rebased onto master for Conda.jl build fix.

ChrisRackauckas:
2 of them worked. @KirillZubov can you do a few last fixes here?

KirillZubov:
how to look at the output of this branch? 
I don't have access, I can only create a new pull request

ChrisRackauckas:
> how to look at the output of this branch?

https://github.com/SciML/SciMLBenchmarks.jl#inspecting-benchmark-results

> I don't have access, I can only create a new pull request

I'll give you access.

ChrisRackauckas:
You should have access to push to PR branches now.

KirillZubov:
ok, thanks. I will fix it

KirillZubov:
@ChrisRackauckas  some error in buildkite `Bypass failed signature check for '.buildkite/launch_benchmarks.yml'?`

ChrisRackauckas:
yeah I've been manually overriding it. @staticfloat do you know why I've been needing to do that?

staticfloat:
This means that someone modified something that deals with secrets, like the part that uploads the benchmarks or something like that.

To avoid drive-by PRs coming and modifying the `pipeline.yml` to just print out all of our secrets, we detect if it's been changed from the previously-known signature, and only allow admins to manually allow the pipeline to run.  To give it the stamp of approval, someone with the private key (right now that's just me and Anand, I think) needs to re-sign the pipelines.  I've done this and pushed the new signature onto `master`, so if you rebase this PR, it should pick it up.

ChrisRackauckas:
Looks like these all worked.

GridTraining isn't normalized to the same loss as the others though, so I'm not sure how fair that comparison ends up being in some sense. It has the caveat that low error in grid training may not be low error against the analytical solution? We should find a way to remedy that, but for now this looks good.

Issue #359, opened 2021-12-11
More DAE benchmarks


ChrisRackauckas:
https://archimede.dm.uniba.it/~testset/testsetivpsolvers/?page_id=26#DAE

- [ ] [Chemical Akzo Nobel](https://archimede.dm.uniba.it/~testset/problems/chemakzo.php)
- [ ] [Andrews' squeezing mechanism problem](https://archimede.dm.uniba.it/~testset/problems/andrews.php)
- [ ] [Transistor amplifier problem](https://archimede.dm.uniba.it/~testset/problems/transamp.php)
- [ ] [Charge-pump problem](https://archimede.dm.uniba.it/~testset/problems/pump.php)
- [ ] [Two bit adding unit problem](https://archimede.dm.uniba.it/~testset/problems/tba.php)
- [ ] [Car axis problem](https://archimede.dm.uniba.it/~testset/problems/caraxis.php)
- [ ] [Fekete problem](https://archimede.dm.uniba.it/~testset/problems/fekete.php)
- [ ] [Slider crank problem](https://archimede.dm.uniba.it/~testset/problems/crank.php)
- [ ] [Water tube problem](https://archimede.dm.uniba.it/~testset/problems/water.php)
- [ ] [NAND gate problem](https://archimede.dm.uniba.it/~testset/problems/nand.php)
- [ ] [Wheelset problem](https://archimede.dm.uniba.it/~testset/problems/wheel.php)

ChrisRackauckas:
If anyone wants to help, what needs to be done is a benchmark file similar to the ROBER DAE one needs to be created:

https://github.com/SciML/SciMLBenchmarks.jl/blob/master/benchmarks/DAE/ROBERDAE.jmd

To do this, you can go to the site and find the benchmark problem description and translate it to the ModelingToolkit.jl form:

https://archimede.dm.uniba.it/~testset/src/problems/andrews.f

If you don't want to setup the solvers, posting the MTK code for the benchmark would be amazing.

caseykneale:
I got a start on the first one, little stuck figuring out the DAE problem, can't find the constructor for that, (yet)
Incase I run out of time(looks like I might) I'll dump the code I have here to maybe save someone some effort:

<details>
  <summary>long code block...</summary>
  
```julia
using ModelingToolkit
using OrdinaryDiffEq
using DiffEqDevTools, ODEInterfaceDiffEq
using LinearAlgebra
using DASSL, DASKR
using Sundials

[...]

PR #372, opened 2021-12-15
Transistor amplifier benchmark

AayushSabharwal:
Added the MTK version of [transistor amplifier problem](https://archimede.uniba.it/~testset/problems/transamp.php), from #359

I'm not sure how the solvers are set up, so everything below line 58 (`tspan = (0.0, 0.2)`) is from ChemicalAkzoNobel.jmd

### DIFF
--- a/benchmarks/DAE/TransistorAmplifier.jmd
+++ b/benchmarks/DAE/TransistorAmplifier.jmd
@@ -1,0 +1,224 @@
+```julia
+using OrdinaryDiffEq, DiffEqDevTools, Sundials, ModelingToolkit, ODEInterfaceDiffEq,
+      Plots, DASSL, DASKR
+using LinearAlgebra
+
+ModelingToolkit.@parameters begin
+      Ub=6.0
+      UF=0.026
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Copy/pasting the solvers is 👍. They should all have about the same setup, though they can be tweaked a little bit to avoid any instabilities that can show up (which messes up the plot), so a tolerance may be tweaked down or specific solvers can be commented out / moved to a separate plot to accommodate that. We'll see what this generates.

AayushSabharwal:
Weird. I'll try running this locally tomorrow and see if I can find the problem.

AayushSabharwal:
Line 56 seems to be the issue, it can't create the system

ChrisRackauckas:
@YingboMa does MTK support parsing linear mass matrices like that right now?

YingboMa:
MTK tries to transform the system so that it is an almost identity matrix, i.e. the semi-explicit form. I tried it on the latest MTK, but initialization of the DAE then fails.

PR #373, opened 2021-12-16
Slider Crank, DAE format

vtjnash:
For #359, from https://archimede.dm.uniba.it/~testset/problems/crank.php

I opted for a fairly direct translation of the code here initially, so that could perhaps be cleaned up a bit later. The solutions graphs here do not match the originals, but possibly show an increasing energy state as phi1 winds up.

@ChrisRackauckas I had to introduce phony variables for D(D(X)) to get the equation to solve (see commit 2). Can you help me figure out also what is wrong with the original formulation?

Unknown-2 svg
Unknown-3 svg
Unknown-4 svg
Unknown-5 svg

### DIFF
--- a/benchmarks/DAE/Crank.jmd
+++ b/benchmarks/DAE/Crank.jmd
@@ -1,0 +1,500 @@
+---
+title: Slider Crank, DAE format
+author: Jameson Nash
+---
+
+```julia
+using OrdinaryDiffEq, DiffEqDevTools, Sundials, ModelingToolkit, ODEInterfaceDiffEq,
+      Plots, DASSL, DASKR
+using LinearAlgebra
+
+# Setup parameter space:
+@parameters(
+    m₁ = 0.36,
+    m₂ = 0.151104,
+    m₃ = 0.075552,
+    l₁ = 0.15,
+    l₂ = 0.30,
+    J₁ = 0.002727,
+    J₂ = 0.0045339259,
+    π  = 3.1415927,
+    EE = 0.20e12,
+    NUE= 0.30,
+    BB = 0.0080,
+    HH = 0.0080,
+    ρ  = 7870.0,
+    γ = 0.0,
+    Ω  = 150.,
+    )
+
+    FACM = ρ*BB*HH*l₂
+    FACK = EE*BB*HH/l₂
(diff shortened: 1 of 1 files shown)


benchmarks/DAE/Crank.jmd
```diff
@@ -0,0 +1,584 @@
...
+```julia
+@named sys = ODESystem(eqs, t)
+simpsys = structural_simplify(sys)
+odaeprob = ODAEProblem(simpsys,[],(0.0,0.1))
+# mmprob = ODEProblem(sys,[],(0.0,0.1))
+daeprob = DAEProblem(sys,ivs,[],(0.0,0.1))
+
+probs = Any[nothing, nothing, nothing]
+refs = Any[nothing, nothing, nothing]
+# mm_sol = solve(mmprob,?,abstol=1/10^14,reltol=1/10^14);
```
ChrisRackauckas, line 412: The lowering of higher derivatives to dummy values occurs in `structural_simplify`, so if you supply an equation with second derivatives you can only solve the simplified form. We should throw an error in this case because we know it's wrong, but 😅 https://github.com/SciML/ModelingToolkit.jl/issues/1389 we'll get to it haha.

ChrisRackauckas: So `ODAEProblem would've been the only one that worked because in the original script that's the only one that used the simplified version of the system.

vtjnash: It could not solve the simpsys either. I had to add dummy variables for both the first and second derivatives.

ChrisRackauckas: Since the problem is given in index 2 form, it might need dummy derivatives for the simplification to give a stable solution, which is something the current simplification technique doesn't do. This might be a good case for @YingboMa to look at.

YingboMa:
I tried it on the latest MTK, but the initialization of the DAE then fails just like https://github.com/SciML/SciMLBenchmarks.jl/pull/372

ChrisRackauckas:
We just merged https://github.com/SciML/SciMLBenchmarks.jl/pull/1459 which was the rebase of this, so 🎉 thanks!

PR #392, opened 2021-12-27
Try to generate doc pages

ChrisRackauckas:


ChrisRackauckas:
@staticfloat What's the right way to do this? I thought it would just copy it over to the benchmarksoutput and then I could turn on gh-pages, but it looks like that's not going to work?

staticfloat:
Where are you actually running the documenter build?

ChrisRackauckas:
Nowhere right now. My plan was to get it sending things to SciMLBenchmarksOutput and then do the build there.

PR #396, opened 2022-01-17
Parallel Extrapolation Methods Benchmarks

utkarsh530:
Testing Parallel Extrapolation Methods on Stiff ODEs.

TODO:
- [ ] Alter with order ranges for best performance (Comment: min_order = 3 gives better performance in some cases rather than the default of 2)
- [ ] Play with sequences (I think they are alright default for the best)
- [ ] Threading options (Polyester is best according to my initial testing)

Let me know if there's anything else to check.

@ChrisRackauckas


ChrisRackauckas:
I rebased onto master to fix threads

ChrisRackauckas:
Continue optimizing in new PRs.

PR #398, opened 2022-01-18
Show threads in test

staticfloat:
Also set `--threads=auto` when running a benchmark


staticfloat:
@ChrisRackauckas This fixes the threads (by adding `--threads=auto` to the build script) and the signature, which we were talking about earlier.

ChrisRackauckas:
Thanks!

PR #408, opened 2022-01-25
Jump benchmarks

isaacsas:
Resurrecting and updating https://github.com/SciML/SciMLBenchmarks.jl/pull/298


ChrisRackauckas:
The manifest has to be on v1.6.

isaacsas:
Is it kosher to just edit it to 1.6, or do I need to regenerate it under 1.6 to make sure it is consistent / works ok?

ChrisRackauckas:
Regenerate it because the manifest format is different.

isaacsas:
@Vilin97 I think I've got this fully updated now, and it looks ok locally, but I do  wonder if there is an issue in `DirectCRDirect`. It is still showing the same asymptotic scaling as NSM, and not matching up with Sanft's results where the work plateaus. I added a callback to make it actually stop simulating at 1e8 events, so it should be replicating what he did. I wonder if we need to go to a larger problem size to see the correct asymptotic scaling, though it seems weird we are well-past where he saw it and it hasn't kicked in...

isaacsas:
I giving this a last run locally to check the html output, and will then need to regenerate the Manifest on 1.7 which I don't currently have installed, but otherwise this is hopefully all good finally.

isaacsas:
I'll update the remaining jump benchmarks to work with this manifest/project and Catalyst.

isaacsas:
@Vilin97 I think this is done -- let me know once you've taken a look at the spatial benchmark and then we can merge if tests have passed.

benchmarks/Jumps/Spatial_Signaling_Sanft.jmd
```diff
@@ -0,0 +1,240 @@
...
+
+    # figure out the indices of species EA and EB
+    @unpack EA, EB = rn
+    EAidx = findfirst(isequal(EA), species(rn))
+    EBidx = findfirst(isequal(EB), species(rn))
+
+    # spatial initial condition
+    # initial concentration of 12.3 nM = 12.3 * 1e-3 μM
+    num_molecules = trunc(Int, micromolar_to_invcubicmicrometer(12.3*1e-3) * (domain_len^3))
+    num_nodes = prod(grid.dims)
```
Vilin97, line 70: defining num_nodes the second time

benchmarks/Jumps/Spatial_Signaling_Sanft.jmd
```diff
@@ -0,0 +1,240 @@
...
+# D is the diffusivity in units of (μm)^2 s⁻¹
+function transport_model(rn, N; domain_len = 6.0, D = 1.0, rng = Random.default_rng())
+    # topology
+    h = domain_len / N
+    dims = (N, N, N)
+    num_nodes = prod(dims)
+
+    # Cartesian grid with reflecting BC at boundaries
+    grid = CartesianGrid(dims)
+    grid
```
Vilin97, line 54: This line doesn't do anything, right?

benchmarks/Jumps/Spatial_Signaling_Sanft.jmd
```diff
@@ -0,0 +1,240 @@
...
+        benchmarks = Vector{BenchmarkTools.Trial}(undef, length(algs))
+
+        # callback for terminating simulations
+
+        for (i, alg) in enumerate(algs)
+            name = names[i]
+            println("benchmarking $name")
+            jp = JumpProblem(dprob, alg, majumps, hopping_constants=hopping_constants,
+                             spatial_system = grid, save_positions=(false,false))
+            b = @benchmarkable solve($jp, SSAStepper(); saveat = $(dprob.tspan[2]), callback) setup = (callback = deepcopy($cb)) samples = 10 seconds = 1500
```
Vilin97, line 174: is 1500 seconds enough to reach 10^8 simulation steps?

Vilin97 (changes requested):

benchmarks/Jumps/Spatial_Signaling_Sanft.jmd
isaacsas, line 54:
```suggestion
```

benchmarks/Jumps/Spatial_Signaling_Sanft.jmd
```diff
@@ -0,0 +1,239 @@
...
+
+    # figure out the indices of species EA and EB
+    @unpack EA, EB = rn
+    EAidx = findfirst(isequal(EA), species(rn))
+    EBidx = findfirst(isequal(EB), species(rn))
+
+    # spatial initial condition
+    # initial concentration of 12.3 nM = 12.3 * 1e-3 μM
+    num_molecules = trunc(Int, micromolar_to_invcubicmicrometer(12.3*1e-3) * (domain_len^3))
+    num_nodes = prod(grid.dims)
```
isaacsas, line 69:
```suggestion
```

benchmarks/Jumps/Spatial_Signaling_Sanft.jmd
isaacsas: When I went up to a mesh of N=120 I only needed ~60 seconds for the median, so ~600 seconds for 10. So I think it should be enough. That said, I'll bump it to an hour.

isaacsas: And of course that is on my laptop; I have no idea how it will perform on the server.

benchmarks/Jumps/Spatial_Signaling_Sanft.jmd
```diff
@@ -0,0 +1,238 @@
...
+        benchmarks = Vector{BenchmarkTools.Trial}(undef, length(algs))
+
+        # callback for terminating simulations
+
+        for (i, alg) in enumerate(algs)
+            name = names[i]
+            println("benchmarking $name")
+            jp = JumpProblem(dprob, alg, majumps, hopping_constants=hopping_constants,
+                             spatial_system = grid, save_positions=(false,false))
+            b = @benchmarkable solve($jp, SSAStepper(); saveat = $(dprob.tspan[2]), callback) setup = (callback = deepcopy($cb)) samples = 10 seconds = 1500
```
isaacsas, line 172:
```suggestion
            b = @benchmarkable solve($jp, SSAStepper(); saveat = $(dprob.tspan[2]), callback) setup = (callback = deepcopy($cb)) samples = 10 seconds = 3600
```

isaacsas:
@Vilin97 thanks for taking a look! For future reference, feel free to use the code suggestion mechanism for small changes like removing the redundancies. Then I can just commit them directly from your comment without having to edit anything locally.

@ChrisRackauckas I think this is good to go now. I am running the spatial benchmark today locally if you want to wait until I get confirmation it finishes properly (I have run it up to N=120 mesh voxels, but not at the largest size in the PR, 240).

ChrisRackauckas:
Yeah I've been following this. It looks good. There is still a depwarn in there but I'm not sure where that's coming from.

isaacsas:
The package version issue doesn't come up locally for me. Could it be because I'm running on a 1.8 rc? Maybe that causes issues with the manifest (though the warning is about packages loaded from the SciMLBenchmarks project file)?

This issue
```julia
The variable syntax (G[1:2N])(t) is deprecated. Use (G(t))[1:2N] instead.
```
is coming from the problem library I think, which has not yet been updated for this depreciation (since it is a recent change in Symbolics).

This one
```julia
┌ Warning: Weave: `options` key is deprecated. Use `weave_options` key instead.
```
seems to be from this PR -- I can update the jmd files with the new notation when I get to work a bit later this morning.

Vilin97:
Is this being worked on? Do we know why it's failing?

isaacsas:
It’s finished. Tests just failed from the server job timeout being too short I believe.

ChrisRackauckas:
Yes, 16h and 40minutes is 1000 minutes which is the time out time. Let's merge this so most of them are updated, and then just update the long one separately. I think on its own it doesn't hit the limit, but all jump benchmarks together does hit the limit.

isaacsas:
FWIW, everything passes and builds locally for me.

PR #411, opened 2022-03-16
[WIP][Don't Merge]Improvements in Parallel Extrapolation Benchmarks

utkarsh530:
I made this PR to try things with `RFLUFactorization` and other general comments.


utkarsh530:
I want to point out a regression when testing locally the benchmarks.
On BK:
https://julialang-buildkite-artifacts.s3.amazonaws.com/scimlbenchmarks/06dadfecca3a96c5c03911876fdef818ee88aaefaa296d33d7ca86732a5298cf/html/StiffODE/Pollution.html

download

On Locally:
Pollution_14_1
[POLLU Work-Precision Diagrams.pdf](https://github.com/SciML/SciMLBenchmarks.jl/files/8274734/POLLU.Work-Precision.Diagrams.pdf)

The version info and `Project.toml` and `Manifest.toml` can be checked for any version differences

ChrisRackauckas:
It's probably because the server has a lot more cores and a lot more threads. To fill the threads, you'll want to use a lot higher orders on the server.

ChrisRackauckas:
You're not getting rid of compilation before profiling. You need to run the equation once before profiling. Then in the profile, you need to profile multiple solves, i.e. `@profile for i in 1:1000 ... end` in order to get an accurate profile

utkarsh530:
I have done that now. You can open SVG separately in browser for interactive viewing

ChrisRackauckas:
Screenshot 2022-03-22 033532

The majority of the time is in the `alg_cache` just building the setup arrays. 

https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/caches/extrapolation_caches.jl#L1198
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/caches/extrapolation_caches.jl#L1208
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/caches/extrapolation_caches.jl#L1231

There then is not enough solving to recover this cost. This suggests that what extrapolation really needs is longer / more costlier ODEs

```
Environment:
  BUILDKITE_PLUGIN_JULIA_CACHE_DIR = /cache/julia-buildkite-plugin
  JULIA_DEPOT_PATH = /cache/julia-buildkite-plugin/depots/5b300254-1738-4989-ae0a-f4d2d937f953
```

@staticfloat I don't know if we're setting a thread count for Julia? Could it be an issue due to how weave is launched? `versioninfo()` usually shows the thread count, i.e. locally:

```julia
julia> versioninfo()
Julia Version 1.7.2
Commit bf53498635 (2022-02-06 15:21 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: AMD Ryzen 9 5950X 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, znver3)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 16
```

so it looks like we're not getting multithreading in our benchmarks for this reason.

utkarsh530:
I get that; we know that the problem is typically small for multi-threading gains. However, I am trying to shed light on the different issues, i.e. regression in the benchmarks on Buildkite on most of the algorithms. You can see from the plot I've attached. So I am not sure from that regression is arising? Maybe some config of the machine or Buildkite? See the benchmark result here as well:

On `deepsea2`:

```
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  132.488 μs …  26.103 ms  ┊ GC (min … max):  0.00% … 96
.55%
 Time  (median):     140.968 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   164.569 μs ± 713.006 μs  ┊ GC (mean ± σ):  12.01% ±  2
.76%

        ▁▅▇█▇▆▃▂▂                                                
  ▁▁▁▃▄▇██████████▇▇▇▆▅▆▅▅▄▃▃▂▂▂▂▂▃▃▃▄▄▅▆▇▆▆▆▇▇▆▅▆▅▄▄▄▄▃▃▃▂▂▂▂▁ ▄
  132 μs           Histogram: frequency by time          161 μs <

 Memory estimate: 222.58 KiB, allocs estimate: 711.
```

On BK:
```
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  308.147 μs … 35.855 ms  ┊ GC (min … max):  0.00% … 96.
73%
 Time  (median):     318.077 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   419.413 μs ±  1.473 ms  ┊ GC (mean ± σ):  19.01% ±  5.
35%

     ▁█▇                                                        
  ▂▃▅████▅▄▃▃▃▃▃▃▂▂▂▃▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▁▂▂▂▂▂▃▃▃▃▃▄▅▅▄▄▃▃▂▂▂▂▂▂▂▂ ▃
  308 μs          Histogram: frequency by time          406 μs <

 Memory estimate: 699.30 KiB, allocs estimate: 1600.
```
Here's the profile when the method is run in `deepsea2`
profile

ChrisRackauckas:
Expand it like I did. My guess is that a lot less time is spent in the allocation of the caches.

ChrisRackauckas:
> Maybe some config of the machine or Buildkite?

Yes, I just mentioned BK seems to not be multithreading.

utkarsh530:
image

utkarsh530:
> Yes, I just mentioned BK seems to not be multithreading. It may still be allocating per-thread though.

Would that multi-threading issue affect `Rosenbrock23` as well? I think they have also gone up.

ChrisRackauckas:
Yes, because it uses polyester threading in the lu-factorization, but not if there's no Julia threads.

utkarsh530:
> The majority of the time is in the alg_cache just building the setup arrays.

Considering that a good chunk of time is being taken up there, would it be good if we build a single dimension for non-threaded cases? We are building `Threads.nthreads()` size arrays every time but we use only `W[1]` of it. For eg see here:
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/perform_step/extrapolation_perform_step.jl#L2785

ChrisRackauckas:
> Considering that a good chunk of time is being taken up there, would it be good if we build a single dimension for non-threaded cases?

Definitely. It should always be using `alg.threads` instead of `Threads.nthreads()`.

staticfloat:
For the record, on buildkite we set `JULIA_CPU_THREADS` and limit the number of CPUs you're allowed to use through cgroups.  You have to use `--threads=auto` to use multithreading (as usual) so make sure you pass that argument to Julia to actually start multithreaded.  Different machines have differing numbers of threads available; e.g. "normal" runners only have 16 threads, while the `exclusive` runners have 128 available.

ChrisRackauckas:
It is doing auto threads I thought though?

https://github.com/SciML/SciMLBenchmarks.jl/blob/master/.buildkite/run_benchmark.yml#L48

staticfloat:
Yeah, agreed.  So what is your evidence for saying that it's not multithreading?

ChrisRackauckas:
Look at the bottom:

```julia
Julia Version 1.6.5
Commit 9058264a69 (2021-12-19 12:30 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD EPYC 7502 32-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, znver2)
Environment:
  BUILDKITE_PLUGIN_JULIA_CACHE_DIR = /cache/julia-buildkite-plugin
  JULIA_DEPOT_PATH = /cache/julia-buildkite-plugin/depots/5b300254-1738-4989-ae0a-f4d2d937f953
```

If Julia had access to threads, it would be in the environment.

staticfloat:
I'm not sure what you mean?

```
$ julia-1.6 --threads=auto
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.6.5 (2021-12-19)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> versioninfo()
Julia Version 1.6.5
Commit 9058264a69 (2021-12-19 12:30 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD Ryzen 9 3950X 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, znver2)
Environment:
  JULIA_PKG_PRECOMPILE_AUTO = 1
  JULIA_PKG_SERVER = https://pkg.julialang.org

julia> Threads.nthreads()
32
```

ChrisRackauckas:
Oh... yeah... you don't have to.... do... environment.... 😮‍💨

PR #413, opened 2022-04-04
Migrate to Franklin

pakzy:


ChrisRackauckas:
I don't see how this is connected to the BuildKite. And wouldn't this go on https://github.com/SciML/SciMLBenchmarksOutput?

PR #425, opened 2022-05-24
Add MethodOfLines Benchmark

xtalax:
Adds a benchmark for [MethodOfLines.jl](https://github.com/SciML/MethodOfLines.jl), using Burgers Equation.


ChrisRackauckas:
@staticfloat or @anandijain , the hash thing is back?

staticfloat:
I fixed a couple problems in https://github.com/SciML/SciMLBenchmarks.jl/pull/426, merge that then rebase this on top of `master`.

benchmarks/MOLPDE/burgers_fdm_wpd.jmd
```diff
@@ -48,17 +49,76 @@ function burgers(N, ϵ)
...
+    x_max = dx*N
+    t_max = 1.0
+    
+    xs = dx:dx:x_max
+
+    eq = Dt(u(x, t)) ~ ϵ * Dxx(u(x, t)) - u(x, t) * Dx(u(x, t))
+
+    bcs = [u(x,0.0) ~ exp(-(x - μ0)^2 / (2 * σ0^2)),
+           u(dx,t) ~ 0,
+           d(dx*N,t) ~ 0]
```
MilkshakeForReal, line 76: `u(dx*N,t)~0`

benchmarks/MOLPDE/burgers_fdm_wpd.jmd
```diff
@@ -48,17 +49,76 @@ function burgers(N, ϵ)
...
+# Time discretization with MethodOfLines
+println("Discretization time wih MethodOfLines.jl:")
+@time xs, prob_mol = burgers_mol(512, 1e-3)
+
+# Solve with MOL
+sol_mol = solve(prob, Vern9(); abstol=1e-14, reltol=1e-14)
+test_sol_mol = TestSolution(sol_mol);
+
 tslices = [0.0 0.25 0.50 0.75 1.00]
+labels = ["t = $t" for t in tslices
```
MilkshakeForReal, line 111: `labels = ["t = $t" for t in tslices]`

ChrisRackauckas:
Let's get this updated. See https://benchmarks.sciml.ai/dev/StiffODE/Bruss/. We can now have multiple problems and allow for a `:prob_choice` in the benchmark. It would be interesting to update this to be alternative prob choices on the FDM ones and see how MOL.jl performs against the hand-written discretizations.

xtalax:
I already have the working bruss code, let me see how to integrate this

xtalax:
I expect `WorkPrecisionSet`  will need special handling for pdesol, I'm just passing the original_sol at the moment. I will duplicate all the MTKitize algs as that's the most equivalent

xtalax:
@ChrisRackauckas  how is bruss looking now? And will unused prob kwargs be passed down to the solve, or do I need to do something special?

ChrisRackauckas:
Add it as a 4th version in Bruss. Don't delete the fastprob one.

xtalax:
@ChrisRackauckas looking at the wp setups, there are cases where `:prob_choice` is 2 for 2 consecutive identical setups. Should these duplicated ones be changed to 3?

ChrisRackauckas:
Most of the different problem checking is done in just the single test. That is because that's for problem formulation, while the other tests of solvers. However, there is the confounding fact that GMRES benefits a lot less from having a fast symbolic Jacobian definition than the others (because it doesn't use it), and so the balance between solvers can be dependent on the problem formulation. For that reason, there's some cases where the modelingtoolkit version is included in the solver tests to showcase this effect, and notably that uses `jac=true, sparse=true`, and that's why you see a major difference with that included (specifically with the KLU version).

ChrisRackauckas:
Failures look like typos

xtalax:
Looks like the fails were caused by MethodOfLines not being in the manifest, I have instantiatec

ChrisRackauckas:
Rebase for v1.8

xtalax:
Rebased

ChrisRackauckas:
need to update the manifest

ChrisRackauckas:
In the manifest:

> julia_version = "1.8.1"

xtalax:
You mean this should be 1.8.2, or omitted? There are examples of both styles across the package

ChrisRackauckas:
it should be v1.8.2 to match the CI machine. I think that's why it's greyed out and I cannot unblock the runs.

ChrisRackauckas:
There's failures in a good chunk of these.

xtalax:
Does WPSet pass down kwargs to the solve? Need to set `wrap=Val(false)`

ChrisRackauckas:
It does. If it's a problem-specific thing, then the setup will too.

benchmarks/StiffODE/Bruss.jmd
```diff
@@ -357,7 +413,7 @@ setups = [
           ]
 names = ["KenCarp47 KLU MTK", "KenCarp47 GMRES MTK",
          "FBDF KLU MTK", "FBDF GMRES MTK",
-         "CVODE MTK KLU", "CVODE iLU MTK GMRES"
```
ChrisRackauckas, line 360: Why this change?

benchmarks/StiffODE/Bruss.jmd
```diff
@@ -400,7 +456,7 @@ setups = [
 names = ["KenCarp47 KLU MTK", "KenCarp47 GMRES MTK",
          "FBDF KLU MTK", "FBDF GMRES MTK",
          "Rodas5P GMRES MTK",
-         "CVODE MTK KLU", "CVODE iLU MTK GMRES"
```
ChrisRackauckas, line 403: why this change?

xtalax: Because that setup doesn't reference problem 2

benchmarks/StiffODE/Bruss.jmd
ChrisRackauckas: change the prob choice not the title here.

ChrisRackauckas:
Builds failed.

ChrisRackauckas:
@xtalax status here?

xtalax:
Status is that I can't work out why this is still failing, could copy out the code and project and try to run manually to debug

xtalax:
@ChrisRackauckas this should be done now

ChrisRackauckas:
Okay let's continue the MethodOfLinesPDE set in a later PR

PR #438, opened 2022-07-24
Fix some benchmarks

utkarsh530:


utkarsh530:
The results in Julia v1.7.2 are better for the methods: https://github.com/utkarsh530/ParallelExBenchmarks.jl/tree/main/plots

PR #450, opened 2022-07-26
bump nbody

ChrisRackauckas:


ChrisRackauckas:
@SebastianM-C could you take a look at this? Do you expect it to take more than 10 hours?

SebastianM-C:
Hmm, it could be due to the stepsize calibration for the symplectic solvers. I'll take a look and double check that.

SebastianM-C:
I've re-ran the timestep calibration code
https://github.com/SciML/SciMLBenchmarks.jl/blob/a7df7e7c0fe27c2915769312a8cbd27608c90043/benchmarks/NBodySimulator/liquid_argon.jmd#L123-L129
and I get
```julia
9-element Vector{Float64}:
  1.0
  2.9735980871351813
  3.0005619459841046
  3.0164920767484356
  4.985990322613598
  6.9410438162718835
  8.011579310978117
 18.053161246947383
 35.93506077349945
```
as opposed to the hardcoded
```julia
c_symplectic = [
    1.00,   # VelocityVerlet
    1.05,   # VerletLeapfrog
    0.98,   # PseudoVerletLeapfrog
    1.02,   # McAte2
    2.38,   # CalvoSanz4
    2.92,   # McAte5
    3.74,   # Yoshida6
    8.44,   # KahanLi8
    15.76   # SofSpa10
]
```

I'm not sure why the difference is so big, but it could be either that we have longer simulation times or lower number of steps. Either way, we are now passing lower `c_symplectic` values than it would be needed, and since `dt=τ*cₐ`
https://github.com/SciML/SciMLBenchmarks.jl/blob/a7df7e7c0fe27c2915769312a8cbd27608c90043/benchmarks/NBodySimulator/liquid_argon.jmd#L113
we get smaller timesteps for the symplectic integrators, which are by far the most time consuming part of the benchmark. We give roughly half the value that I obtained, so I think that could double the simulation time (assuming linear scaling). Since this took more than a couple of hours already, I would assume that it could be an explanation for the 10h+ runtime.
I'll try to look to the absolute runtimes from the past and what I get now over the next days.

ChrisRackauckas:
any update?

SebastianM-C:
In the past ~6 hours I've been running `run_benchmark!(results, t, symplectic_integrators, τs[1], c=c_symplectic)`, i.e. just compute the results for the first timesteps and it looks like the runtimes are larger. For example the progress bar for `CalvoSanz4` indicates around 1h 7min remaining (and it started a while ago, it says 32%) and the time in https://benchmarks.sciml.ai/html/NBodySimulator/liquid_argon.html for that one is 42.344 mins.

Even if we ignore the slowdown, the original benchmark took more than 10 hours, just by looking at the `runtime` column on the "fixed simulation time and variable timestep" benchmark. I recall that I ran it partially on a server for several hours (with less than 9 integrators) and just for the PR I enabled all of them. We have 9 integrators that go through 10 sets of timesteps and with the first 2 we can say that we have an average runtime of 40mins. This already gives us a minimum runtime of 12h.

The solution would be to reduce the total simulation time (the `t` variable) and maybe also to recalibrate `c_symplectic` and use larger time steps in `τs`.

ChrisRackauckas:
I just set the max really really high. Let's see how that goes.

PR #451, opened 2022-07-26
bump adaptive sde

ChrisRackauckas:


ChrisRackauckas:
If you instantiate a manifest and then use distributed, do the nodes not have access to those packages?

ChrisRackauckas:
Fixed in https://github.com/JuliaLang/julia/pull/43270 (v1.8)

PR #453, opened 2022-07-28
Bump NeuralPDE v5 benchmarks

ChrisRackauckas:


KirillZubov:
@ChrisRackauckas could you please check.

ChrisRackauckas:
It looks like the two that built look good!

KirillZubov:
why only two were built?

ChrisRackauckas:
I'm not sure, maybe it just hit a compute limit. Let's merge and see. All 5 of the error ones worked, but only 2 of the optimization ones.

PR #473, opened 2022-08-18
change sigmoid to tanh in PINNs' benchmarks

MilkshakeForReal:
The sigmoid function is suboptimal, especially when the inputs are all positive, it leads to all positive gradients. 

See https://rohanvarma.me/inputnormalization/ for details.

It shouldn't change the conclusions drawn by the current benchmarks, but I believe the losses will drop overall.


ChrisRackauckas (approved):
yes

Issue #484, opened 2022-08-22
Fixing the Stan binary for the Bayesian estimation


ChrisRackauckas:
The Stan integration for Bayesian estimation broke. The latest one that worked was v1.4.

https://benchmarks.sciml.ai/stable/ParameterEstimation/DiffEqBayesFitzHughNagumo/
https://benchmarks.sciml.ai/stable/ParameterEstimation/DiffEqBayesLotkaVolterra/

The latest attempt was in https://github.com/SciML/SciMLBenchmarks.jl/pull/362 . If you check the artifacts of the latest build (https://buildkite.com/julialang/scimlbenchmarks-dot-jl/builds/904#0182b81c-d678-4f82-aaf3-e1a49cf5b0b5), you see:

```
Error: IOError: cd(""): no such file or directory (ENOENT)
```

for the call `@btime bayesian_result_stan = stan_inference(prob_ode_fitzhughnagumo,t,data,priors; print_summary=false, vars=(DiffEqBayes.StanODEData(), InverseGamma(2, 3)))`. @Vaibhavdixit02 tried to get the Stan installation to be automatic, but it seems to be pretty hard.

Pinging all who were involved there: @Vaibhavdixit02 , @ven-k , @goedman, @staticfloat. And @mohamed82008 + @storopoli

ChrisRackauckas:
On-going developments are setup to occur in https://github.com/SciML/SciMLBenchmarks.jl/pull/485.

storopoli:
I think that 6 months ago, the Stan devs managed to get an anaconda packaged up for CmdStan and also they are commmited to maintaining it: https://anaconda.org/conda-forge/cmdstan

@Vaibhavdixit02, that could be a good alternative. What do you think?

storopoli:
Additionally, for [`StanSample.jl`](https://github.com/StanJulia/StanSample.jl), we need to export the `CMDSTAN` or `JULIA_CMDSTAN` environment variables as described in the `README.md`:

```bash
export CMDSTAN=your/absolute/path/to/cmdstan
```

```julia
ENV["CMDSTAN"]="_your absolute path to cmdstan_"
```

storopoli:
Looking here: https://github.com/SciML/SciMLBenchmarks.jl/blob/master/.buildkite/run_benchmark.yml#L49

I think that we are exporting the wrong variable name and also the relative path instead of the "absolute path to the cmdstan".

[...]

PR #485, opened 2022-08-22
Attempt another Bayesian inference build

ChrisRackauckas:
Tries to fix https://github.com/SciML/SciMLBenchmarks.jl/issues/484


goedman:
Will take a look after next weekend (traveling abroad without my laptop).

Vaibhavdixit02:
I think we'll wait for @goedman, no point in wasting more CI hours 😢

ChrisRackauckas:
Thanks! Now we have something to start optimizing! We should change the Turing ODEs to use static arrays.

storopoli:
Great! Thanks for the effort!

Vaibhavdixit02:
Awesome! :tada:

PR #487, opened 2022-08-30
Add Gupta model benchmarks

TorkelE:
Add Gupta model benchmarks


ChrisRackauckas:
Unblocked.

The build result shows artifacts which has the text and plots

TorkelE:
It seems to be blocked again. Is there some way for me to unblock it, or does it require someone with permission?

ChrisRackauckas:
Great stuff, thanks. @utkarsh530 can you take a pass at these to add the extrapolation methods?

utkarsh530:
Yeah, made #491

PR #491, opened 2022-09-02
Add extrapolation benchmarks to new bio models

utkarsh530:


ChrisRackauckas:
These all diverged?

utkarsh530:
Sorry for the delayed response:
No, the `QNDF` and `ImplicitHairerWannerExtrapolation` were giving warnings.
See the plots:
Multistate.net:

image

multisite2.net:
image

ChrisRackauckas:
In the first one, CVODE is clearly not a good source of truth according to those plots. We should switch it to something like Rodas5P

PR #499, opened 2022-09-06
bump a few packages

ChrisRackauckas:


ChrisRackauckas:
Merging since these have known fixes

PR #504, opened 2022-09-09
update to v1.8 and update Bio to have nonstiff stuff

ChrisRackauckas:


ChrisRackauckas:
@YingboMa can you recreate this segfault?

TorkelE:
I have started looking at stuff. Only gotten so far that I am getting the crash as well whenever the system is big enough.

PR #519, opened 2022-09-27
update nonstiff dde

ChrisRackauckas:


ChrisRackauckas:
@devmotion can you help update the naming here?

devmotion:
Can you explain what you mean with naming? It seems you already updated the dependencies and import statements?

ChrisRackauckas:
If you look at the artifacts, it looks like there are errors because the problem names changed. IIRC there was some PR you did that moved a few around.

devmotion:
Ah I forgot that we had removed the deprecated names. Hopefully it should be fixed by https://github.com/SciML/SciMLBenchmarks.jl/pull/519/commits/e2876a1e04040e6b93284c9ca5860a9bc4dd6bcf.

PR #522, opened 2022-09-27
add fast bruss implemenation

ChrisRackauckas:


xtalax:
I'll wait for this before adding MOL to the WorkPrecisionSets as I'd need to change the prob numbers else

PR #527, opened 2022-10-08
Update Bio WP diagrams

TorkelE:
Updates the WP diagrams for the bio models.

PR #531, opened 2022-10-26
Add small bio models

TorkelE:
Contains the changes to the 3 smaller models (multistate, multisite2, and egfr_net) from https://github.com/SciML/SciMLBenchmarks.jl/pull/527

This would enable us to fix these ones while figuring out the long run times of the other models.


TorkelE:
No longer needed.

PR #535, opened 2022-11-23
full specialize in Bidkhori2012

ChrisRackauckas:


ChrisRackauckas:
@utkarsh530 watch this one. If it fails we need to figure out what regressed.

ChrisRackauckas:
Yeah something happened with an implicit extrapolation thing. @utkarsh530 would be good to fix it, I assume it has a quick fix.

utkarsh530:
I think the error was before as well (Sparse Jacobian)? The automatic and analytic jacobian passed the tests, though, as usual.
https://benchmarks.sciml.ai/stable/Bio/Bidkhori2012/

ChrisRackauckas:
oh it doesn't have sparse support? Okay, open an issue on OrdinaryDiffEq and we should cut it out of this benchmark.

PR #558, opened 2022-12-24
Bump stiff Bio models

ChrisRackauckas:


TorkelE:
This got canceled, have I missed something?

ChrisRackauckas:
The MIT computers are down, so this is waiting on @vchuravy and @staticfloat to get something back up.

TorkelE:
Right, think I heard but didn't really think of the implications

TorkelE:
I think in the BCR file, on line 136,
```julia
preccache[] = ilu(W, τ = τ1)
```
should be changed to 
```julia
preccache[] = ilu(W, τ = τ)
```

I tried to update, but got an access denied when I tried to push.

benchmarks/Bio/BCR.jmd
```diff
@@ -114,13 +114,14 @@ end
 #### Declare pre-conditioners
 ```julia
 using IncompleteLU, LinearAlgebra
+const τ = 1e2
+const τ2 = 1e2
 
 jaccache = sparsejacprob.f.jac(oprob.u0,oprob.p,0.0)
 W = I - 1.0*jaccache
-prectmp = ilu(W, τ = 50.0)
+prectmp = ilu(W, τ = τ1)
```
ChrisRackauckas, line 122:
```suggestion
prectmp = ilu(W, τ = τ)

```

PR #565, opened 2023-01-13
Adds Coevolve aggregator to the Jumps benchmark suite

gzagatti:
This PR adds the new `Coevolve` aggregator (added in PR SciML/JumpProcesses.jl#276) to the Jumps benchmark suite.


ChrisRackauckas:
Does this method make sense in these benchmarks? Since it's all constant rates, it's just going to have overhead compared to the other techniques? Is this just to measure that overhead?

gzagatti:
The benchmark is to measure the overhead. When passed constant rates, `Coevolve` should obtain a performance very close to `NRM` since the algorithm reduces to the latter.

ChrisRackauckas:
The benchmarking machine had a hard drive failure. Remind me to make this rerun after the hardware is fixed up

PR #573, opened 2023-01-29
Update bio benchmarks with comparisson

TorkelE:
This will add a final WP plot at the end of each of the Bio benchmarks files. 

The idea is that for the Catalyst paper, we can just download the figures from the SciMLBenchmakrs docs directly (since they only compare within Catalyst, unlike the main figure where we compare between tools). So I am basically adding proposed paper supplementary figures at the end here.

For each summary plot, I basically pick the best benchmarks across the docs. In some cases I also try and pick the best from each set (unless where a set only have really bad ones).

### DIFF
--- a/benchmarks/Bio/BCR.jmd
+++ b/benchmarks/Bio/BCR.jmd
@@ -34,13 +34,13 @@ obs = [eq.lhs for eq in observed(rn)]
 show(to)
 
 tspan = (0.,tf)
-@timeit to "ODEProb No Jac" oprob = ODEProblem{true, SciMLBase.FullSpecialize}(osys, Float64[], tspan, Float64[])
+@timeit to "ODEProb No Jac" oprob = ODEProblem(osys, Float64[], tspan, Float64[])
 show(to)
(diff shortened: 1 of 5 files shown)


TorkelE:
Seems a commit "`Merge pull request #4 from SciML/master`" snuck in. I think that was when I tried to ensure that my version of SciMLBecnhmarks (TorkelE/SciMLBenchmarks) was up-to-date with the main repo. Might this become a problem?

TorkelE:
Before finalizing this I will need to figure out https://github.com/SciML/DiffEqDevTools.jl/issues/103 as well.
(I know that different markers are currently auto generated. But for clarity I have the same markers for the different solvers in the paper, so would be nice to use the same marker for e.g. `QNDF` in the other figures as in the WP diagrams)

ChrisRackauckas:
> I think that was when I tried to ensure that my version of SciMLBecnhmarks (TorkelE/SciMLBenchmarks) was up-to-date with the main repo. Might this become a problem?

Yes, this is a problem. You're ignoring my code changes. Please rebase

TorkelE:
Thanks, closing this while I try to figure out.

PR #576, opened 2023-02-11
add bio benchmarks summary plot

TorkelE:
Adds summary plots for the 5 bio benchmarks (to be used in Catalyst paper).

### DIFF
--- a/benchmarks/Bio/BCR.jmd
+++ b/benchmarks/Bio/BCR.jmd
@@ -34,13 +34,13 @@ obs = [eq.lhs for eq in observed(rn)]
 show(to)
 
 tspan = (0.,tf)
-@timeit to "ODEProb No Jac" oprob = ODEProblem{true, SciMLBase.FullSpecialize}(osys, Float64[], tspan, Float64[])
+@timeit to "ODEProb No Jac" oprob = ODEProblem(osys, Float64[], tspan, Float64[])
 show(to)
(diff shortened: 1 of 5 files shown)


benchmarks/Bio/BCR.jmd
```diff
@@ -34,13 +34,13 @@ obs = [eq.lhs for eq in observed(rn)]
 show(to)
 
 tspan = (0.,tf)
-@timeit to "ODEProb No Jac" oprob = ODEProblem{true, SciMLBase.FullSpecialize}(osys, Float64[], tspan, Float64[])
```
ChrisRackauckas, line 37: No, we should keep this.

benchmarks/Bio/egfr_net.jmd
```diff
@@ -34,12 +34,12 @@ obs = [eq.lhs for eq in observed(rn)]
 show(to) 
 
 tspan = (0.,tf)
-@timeit to "ODEProb No Jac" oprob = ODEProblem{true, SciMLBase.FullSpecialize}(osys, Float64[], tspan, Float64[])
```
ChrisRackauckas, line 37: why are these removed in each?

TorkelE:
Not sure, I only ever added a bit at the end and changed nothing else. My fork must never have contained the `{true, SciMLBase.FullSpecialize}` in the first place. Not sure why because I just re-forked it to ensure it was new. Will fix.

PR #582, opened 2023-03-01
Add ESDIRK methods benchmarks

ErikQQY:


ChrisRackauckas:
You'll need to also update the manifest.

ErikQQY:
When I update the manifest, I encountered a compatibility issue, I tracked it down then I think maybe it is the ```SparsityDetection``` package restriction caused the error:

image

Since we already used Symbolics.jl, I think we can remove this package?

ChrisRackauckas:
yes remove it

ChrisRackauckas:
The new schemes may have some issue in their adaptivity or something. I'd isolate it down to smaller cases.

PR #584, opened 2023-03-07
Adds Multivariate Hawkes benchmark.

gzagatti:
This PR adds the Multivariate Hawkes benchmark for comparing jump algorithms that accept variable rates.

The document contains a section that produces QQ-plots which are useful to check the correctness of the simulations before benchmarking them.

The benchmark also benchmarks a Python library. The document installs and sets up the required Conda environment. However, I don't know whether that is the recommended way to do that.

The document is quite long as I aimed to keep it self-contained. I am looking forward to some feedback on how to improve it.


gzagatti:
It looks like that despite the fact that the Build Kite processes finished successfully, [the markdown file](https://julialang-buildkite-artifacts.s3.amazonaws.com/scimlbenchmarks/38731958b3c009930437d2a8fec1d238301fac20aad757feb52b011cdf029221/markdown/Jumps/MultivariateHawkes.md) did not compile correctly.

```
...
Error: InitError: could not load library "/cache/julia-buildkite-plugin/dep
ots/5b300254-1738-4989-ae0a-f4d2d937f953/conda/3/x86_64/lib/libpython3.10.s
o.1.0"
/cache/julia-buildkite-plugin/depots/5b300254-1738-4989-ae0a-f4d2d937f953/c
onda/3/x86_64/lib/libpython3.10.so.1.0: cannot open shared object file: No 
such file or directory
during initialization of module PyCall
```

It seems like the Conda environment was generated sucessfully but then `PyCall` could not be loaded.

ChrisRackauckas:
Hmm getting Python to work in this might be hard.

gzagatti:
An option is to skip the Python program when running on cloud. What do you think?

ChrisRackauckas:
Let's go with that to start and then improve it in steps.

gzagatti:
PR #585 turns off the Python benchmark on the cloud.

PR #585, opened 2023-03-10
MultivariateHawkes: Do not run Python benchmark in the cloud

gzagatti:
This PR removes the Python benchmark from the cloud as it is currently fails. 

The Python benchmark works locally. To run it locally just set in the file:

```julia
const BENCHMARK_PYTHON = true
const REBUILD_PYCALL = true
```


ChrisRackauckas:
Awesome thanks

PR #597, opened 2023-04-26
CoevolveSynced and Synapse benchmarks WIP

gzagatti:
This adds benchmark for the new aggregator which is under review `CoevolveSynced` (PR SciML/JumpProcesses.jl#315) as well as a new jump model to be benchmarked against, the synapse model.

Discussions on the synapse model can be found in PR rveltz/SynapseElife#1 and the [discourse thread](https://discourse.julialang.org/t/help-me-beat-lsoda/88236).

You probably want to review this PR only once `CoevolveSynced` has been reviewed (PR SciML/JumpProcesses.jl#315) and incorporated into the main branch of `JumpProcesses.jl`


benchmarks/Jumps/Manifest.toml
```diff
@@ -1,8 +1,13 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.8.5"
+julia_version = "1.9.0-rc2"
```
ChrisRackauckas, line 3: The CI is still using v1.8.5 which is what's causing the issue.

gzagatti: Thanks! Let me change that.

ChrisRackauckas:
```
┌ Info: last bit
--
  | │   length(res.xcsol.t) = 3373
  | │   tt[end] = 998.5891927961502
  | └   p_synapse.t_end = 1000.0
  | ┌ Warning: The simulation did not reach requested simulated time.
  | └ @ Synapse /cache/julia-buildkite-plugin/depots/5b300254-1738-4989-ae0a-f4d2d937f953/packages/Synapse/0lKxH/src/Solve.jl:236
```

Lots of them are getting stuck right near the end. Maybe double the allowed time.

gzagatti:
The problem was the saving callback. I think I've fixed it now, let's see.

ChrisRackauckas:
That's odd that some say unstable at the very end and never earlier.

ChrisRackauckas:
Is it only checking for instability at the end?

gzagatti:
Indeed, the code only checks whether the simulation reached the desired simulation time at the end. 

There are three stages in the simulation:
1. t ∈ [0, 500]: Glutamate is off. There are no spikes.
2. t ∈ [500, 501]: Glutamate is on. Spike.
3. t ∈ [501, 1_000]: Glutamate is off. Recovery.

The plot below illustrates one of the variables:


There's a comment on top of the last simulation bit that says:

> reaching tend: we simulate the synapse with Glutamate OFF until simulation end time required by the user. In  most protocol, this is taking most of the time.

The instability only takes place when solving via `PDMP`. But, I am not sure why this happens only in a few of the cases (6 / 50 simulations raise a stability warning, in 2 / 6 warnings the simulation is aborted before reaching the end). I think it's problem specific, might be related with the stiffness of the problem on the recovery bit.

ChrisRackauckas:
I bumped CI so it matches v1.9 as well

gzagatti:
@ChrisRackauckas BuildKite is stuck at `Dynamically launch run_tutorial.yml`. Is there anything we can do about it?

PR #621, opened 2023-08-05
[buildkite] Update S3 credential handling

staticfloat:
We were alerted that our S3 credentials had leaked in a build that dumps
the environment when it errors out [0]. This PR rotates to a new key,
and also wipes out the key in the environment during the build itself,
so that sensitive values won't get written out in debugging information
like that.

[0] https://github.com/SciML/SciMLBenchmarksOutput/blob/8d80b1840228c63e8c6fdc597049211a8b916dea/markdown/DynamicalODE/single_pendulums.md

### DIFF
--- a/.buildkite/build_benchmark.sh
+++ b/.buildkite/build_benchmark.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -euo pipefail
+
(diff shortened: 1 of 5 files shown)


staticfloat:
Looks good now!

ChrisRackauckas:
Thanks!

PR #628, opened 2023-08-14
Create sparse linear solver PDE benchmark

ChrisRackauckas:
CC: @j-fu


j-fu:
Hi, I am thinking about adding two cases with iterative solvers:
Jacobi preconditioned Krylov and  ilu (from ILUZero.jl) preconditioned Krylov. So far the examples are in fact symmetric, so Krylov would mean CG.

As for Jacobi I would expect to see improvement in scaling with _increasing_ dimension - being worst in 1D. This is typical for iterative solvers (besides multigrid an the like). 

ILUZero.ilu0 is an exception to that rule as it seems to be exact in 1D.

PR #629, opened 2023-08-14
Add pardiso

ChrisRackauckas:
@j-fu I left it out of the first one since there seemed to be some installation issues, but trying it now in isolation

### DIFF
--- a/benchmarks/LinearSolve/Manifest.toml
+++ b/benchmarks/LinearSolve/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.9.2"
 manifest_format = "2.0"
-project_hash = "f34d6c06491ace514f6f9bc08e63c12bca59bee3"
+project_hash = "a223202305ec5ecd5db699711e7dcdd85644c64b"
 
 [[deps.ADTypes]]
 git-tree-sha1 = "f5c25e8a5b29b5e941b7408bc8cc79fea4d9ef9a"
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
```
Error: could not load library "libmkl_rt"
libmkl_rt.so: cannot open shared object file: No such file or directory
```

@staticfloat would you know why the JLL doesn't work on this system? MKLPardiso automatically installs its own JLL and so locally this doesn't require any special setup, but for some reason the benchmarking isn't liking this binary.

staticfloat:
No, I'd need more details.  Perhaps have someone try SSH'ing in and running the same thing on that machine?

staticfloat:
Trying this locally, I see `Pardiso` errors inside of `pardisoinit(::MKLPardisoSolver)` for `Pardiso v0.5.0`, but works for `Pardiso v0.5.3+`.  Perhaps you just need to update the benchmark's environment?

The MWE to reproduce the breakage:

```
julia> using Pkg
       Pkg.activate(;temp=true)
       Pkg.add(;name="Pardiso", version=v"0.5.0")
       using Pardiso
       Pardiso.pardisoinit(Pardiso.MKLPardisoSolver())
```

ChrisRackauckas:
Okay I see what was going on now.

```
    Updating `~/.julia/dev/SciMLBenchmarks/benchmarks/LinearSolve/Manifest.toml`
  [46dd5b70] ↑ Pardiso v0.5.0 ⇒ v0.5.4
⌅ [856f044c] ↓ MKL_jll v2023.2.0+0 ⇒ v2022.2.0+0
```

It seems like it's https://github.com/JuliaSparse/Pardiso.jl/pull/91

PR #635, opened 2023-08-24
New sections: Interval rootfinding, nonlinear solvers, BVPs, and optimization

ChrisRackauckas:
Some of them just have stub benchmarks for now.


ChrisRackauckas:
@Vaibhavdixit02 @yash2798 @ErikQQY make note that this will add the benchmark sections we can start populating.

PR #670, opened 2023-08-31
Add optimal power flow benchmark of NLP frameworks

ChrisRackauckas:
@Vaibhavdixit02 @ccoffrin I might need a little help getting this finished, but let's see what the first build says.

@ccoffrin where's the code for building the nice tables for the summary?


benchmarks/OptimizationFrameworks/optimal_powerflow.md
```diff
@@ -0,0 +1,1767 @@
...
+        "time_build" => model_build_time,
+        "time_solve" => solve_time,
+        #"time_callbacks" => TBD,
+    )
+end
+
+solve_opf_nlpmodels("data/pglib_opf_case5_pjm.m")
+```
+
+## Nonconvex
```
Vaibhavdixit02, line 906: I think instead of Nonconvex it should have the NLPModels implementation which seems to be quite performant now.

benchmarks/OptimizationFrameworks/optimal_powerflow.md
```diff
@@ -0,0 +1,1767 @@
+---
+title: Optimal Powerflow Nonlinear Optimization Benchmark
+author: Chris Rackauckas
+---
+
+## Optimization.jl
+
+Constraint optimization implementation reference: https://github.com/SciML/Optimization.jl/blob/master/lib/OptimizationMOI/test/runtests.jl
+Other AD libraries can be considered: https://docs.sciml.ai/dev/modules/Optimization/API/optimization_function/
+However ForwardDiff is the only one that is compatible with constraint functions
```
Vaibhavdixit02, line 10: Except Tracker all other backends now support constraints

ChrisRackauckas: Yes, let's get this working and then merged. Then after merging, let's setup the benchmark to run with multiple Optimization.jl choices.

ChrisRackauckas:
>I had tried to use the forward mode with sparse handling, but it has numerical instability in the hessian calculation (I think) hence it doesn't converge in the same number of steps as the other backends.

It should be exact? There should be no difference between the hessian calculated with sparse and dense with the same autodiff. If you have an example of that failing we should get an MWE.

> . Trying to use Enzyme took up quite a lot of days and I dropped it but now we have zygote and reversediff with the caveat of no sparse which is probably done now after Avik's PR but that didn't have nice API for hessian and hessvec yet so I'll spend some time on that to get it to work (but maybe the dense reversemode is still better than forwardmode?)

Avik's PR automates a lot of it so it shouldn't be very difficult to use. Extending it to Hessians just requires using it on a gradient calculation.

ChrisRackauckas:
Did Enzyme have an issue with this benchmark in particular? I don't see an issue open on Enzyme.jl about this?

Vaibhavdixit02:
>There should be no difference between the hessian calculated with sparse and dense with the same autodiff. If you have an example of that failing we should get an MWE.

Okay, I'll try it again.

>Avik's PR automates a lot of it so it shouldn't be very difficult to use. Extending it to Hessians just requires using it on a gradient calculation.

It's not very trivial but yes it should be not too hard, I talked to him about it will do it.

>Did Enzyme have an issue with this benchmark in particular? I don't see an issue open on Enzyme.jl about this?

It's more of a usability issue, the last error I had was the temporary storage one (known issue) but to work around that it would probably need a lot of rewriting. So that would have to be a separate implementation from the other backends.

Vaibhavdixit02:
Oh also the MTK implementation has the weird issue of the objective value not being exactly the same after `structural_simplify` (it's close but that's not good enough here I suppose 😅)

ccoffrin:
> @ccoffrin where's the code for building the nice tables for the summary?

I don't have a good reference for you. It is fairly specialized to a specific computing setup, but I hope this is useful. The gist is as follows,

This brach of rosseta-opf has a CLI interface which I use to run a bunch of models in parallel, https://github.com/lanl-ansi/rosetta-opf/blob/cli/rosetta-opf-cli.jl

They each output a line "DATA, ..." and I use grep on some standard-out log files to turn these lines into a csv.  Then I use a code like this to render the markdown tables from the CSV files,
```
using DataFrames, Formatting, CSV

HEADER = ["DATA", "case", "variables", "constraints", "feasible", "cost", "time_total", "time_data", "time_build", "time_solve", "time_callbacks"]

datasets = [
    Dict{Symbol,Any}(:name=>"JuMP", :data_file=>"jump_data.csv"),
    Dict{Symbol,Any}(:name=>"NLPModels", :data_file=>"nlpmodels_data.csv"),
    Dict{Symbol,Any}(:name=>"NonConvex", :data_file=>"nonconvex_data.csv"),
    Dict{Symbol,Any}(:name=>"Optim", :data_file=>"optim_data.csv"),
    Dict{Symbol,Any}(:name=>"Optimization", :data_file=>"optimization_data.csv"),
]

julia_solvers = [1,2,3,4,5]

for dataset in datasets
    dataset[:csv] = CSV.File(dataset[:data_file], header=HEADER, stringtype=String)
end

instance_data = Dict{String,Any}()
for dataset in datasets
    data_lookup = Dict{String,Any}()
    for row in dataset[:csv]
        case_id = get_case_id(row.case)
        data_lookup[case_id] = row

        if !haskey(instance_data, case_id)
            num = parse(Int, filter(isdigit, case_id))
            instance_data[case_id] = (case_id=case_id, buses=num, variables=row.variables, constraints=row.constraints)
        end
    end
    dataset[:lookup] = data_lookup
end

instance_ids = sort(collect(Set(keys(instance_data))), by=(x)->instance_data[x].buses)
println(length(instance_ids), " ", instance_ids)

instance_buses = [instance_data[iid].buses for iid in instance_ids]
instance_variables = [instance_data[iid].variables for iid in instance_ids]
instance_constraints = [instance_data[iid].constraints for iid in instance_ids]

df = DataFrame()
df[!,"Case"] = instance_ids
df[!,"Vars"] = instance_variables
df[!,"Cons"] = instance_constraints
for dataset in datasets[julia_solvers]
    local col = get_data_column(instance_ids, dataset[:lookup], :cost, format_string="{:.5e}")
    df[!,dataset[:name]] = col
end

open("rosetta-opf-julia-cost.md", "w") do f
    PrettyTables.pretty_table(f, df, nosubheader=true, tf=PrettyTables.tf_markdown)
end

df = DataFrame()
df[!,"Case"] = instance_ids
df[!,"Vars"] = instance_variables
df[!,"Cons"] = instance_constraints
for dataset in datasets[julia_solvers]
    local col = get_data_column(instance_ids, dataset[:lookup], :time_total)
    df[!,dataset[:name]] = col
end

open("rosetta-opf-julia-time-total.md", "w") do f
    PrettyTables.pretty_table(f, df, nosubheader=true, tf=PrettyTables.tf_markdown)
end
```

ChrisRackauckas:
Well it at least builds. Some of the benchmark code is pretty horrible though so I don't know how this ever passed as a real benchmark. It'll take another round to fix it to be realistic.

PR #672, opened 2023-09-02
Update jump benchmarks

isaacsas:
With recent fixes for `VariableRateJump`s I locally seemed to get much better performance, so it would be nice to rerun the vrj benchmarks now and see any difference.

I've also moved the synapse benchmark to its own folder (it needs a new project and manifest). It was adding a dependency on a fork of what I think will be an unmaintained package -- it is a paper code release -- so didn't seem appropriate to tie together with the other jump benchmarks (for example, it was holding back the Catalyst version from being updated).


ChrisRackauckas:
HybridJumps needs a Project/Manifest.

ChrisRackauckas:
Following up in https://github.com/rveltz/SynapseElife/issues/2 . I think splitting the hybrid jumps out to a separate group is fine anyways. So I'll merge this but we should really keep the thread going to make sure we don't actually lose the hybrid jumps benchmarks.

isaacsas:
It was previously depending of @gzagatti’s fork, so I’m not sure if that means he has a branch that is different than the main repo that was used in the benchmark.

ChrisRackauckas:
Can you make a PR with the right Project.toml/Manifest.toml to revive it? I'm trying to get a few other PDE things revived right now before tagging a new benchmarks version hopefully by the end of the week.

isaacsas:
Yes, I can try to revive it. Might have to check with @gzagatti about what SynapseElife version/branch/fork it needs, but I assume I can just pull what was in the previous Project.toml.

gzagatti:
I ran the Synapse benchmark out of a custom branch, [`coevolve-ii`](https://github.com/gzagatti/SynapseElife/tree/coevolve-ii) as it was unlikely that my modifications would be merged to the main branch of SynapseElife as it was a paper code release.

I also engage in fruitful discussion with @rveltz on [this PR thread](https://github.com/rveltz/SynapseElife/pull/1). Note that the PR was based on a previous custom branch, `coevolve`.

Initially, I wanted to follow the other benchmark examples and have everything in one file. But I realized that the file would be exceedingly long. Just the parameters file has over 600+ lines.

Next, I considered adding a trimmed down version of SynapseElife to the Jump benchmark folder as a package. But then I thought this would not be allowed since no other benchmark has such a dependency.

Finally, I settled for using a fork of a custom branch of the paper code. 

I agree that porting the benchmark to a format that is more maintainable would be desirable. I will follow the discussion here and on rveltz/SynapseElife#2 to provide support.

PR #698, opened 2023-09-10
Refactor and Fix Optimal Powerflow Benchmark

ChrisRackauckas:
1. Make the code canonical. Using a dictionary for indexing a string that is a number `a["$i"]`is not a thing, nor should that ever be a thing. Please use arrays...
2. Refactoring so that the data handling is common. PowerModels.jl should be the thing to actually refactor and not use untyped unhandled structures as its data representation, but to work around that we just fix the data representation before benchmarking.
3. Sets up a test of implementation correctness, though there doesn't seem to be the right functions in JuMP to setup correctness testing.


odow:
> Using a dictionary for indexing a string that is a number a["$i"]is not a thing, nor should that ever be a thing

I think @ccoffrin wrote this to simulate what EE people would normally write. Domain people want named variables, and having to keep array indices in order leads to bugs.

> PowerModels.jl should be the thing to actually refactor and not use untyped unhandled structures as its data representation

@ccoffrin can comment, but I think the design is intentional. Data handling is not the bottleneck for these problems.

odow:
> though there doesn't seem to be the right functions in JuMP to setup correctness testing.

I've thought a bit about this, but I don't have a good solution. One problem is that JuMP doesn't have the notion that an optimization problem is `min f(x); g(x) <= 0`, so there isn't really a vector of inputs and a vector of constraint functions.

ccoffrin:
> @ccoffrin can comment, but I think the design is intentional. Data handling is not the bottleneck for these problems.

Yes, the untyped Dict in PowerModels is a design choice. There are some use cases I know of where the memory overhead is prohibitive, but runtime performance has not been issue as far as I know.  The runtime of the optimization algorithm is the primary bottleneck in the vast majority of use cases we encounter.

odow:
The untyped data inputs did annoy @jd-lara. But I think they wrote a bunch of conversion code in PowerSimulations.jl.

Keeping a simple data structure (just dicts and vectors) seems important for people coming to PowerModels from MATLAB etc.

odow:
Without understanding the details, it seems like PowerModels might need
https://github.com/lanl-ansi/PowerModels.jl/pull/885

ChrisRackauckas:
> Domain people want named variables, and having to keep array indices in order leads to bugs.

Sure, but use a `Dict{Int,Int}` in that case, or a `SparseVector`. I think making it into `"1"` in order to then put it into a dictionary is pretty unnecessary.

> The runtime of the optimization algorithm is the primary bottleneck in the vast majority of use cases we encounter.

The majority of the runtime in the optimization algorithms is constructing strings for putting into dictionaries.

ChrisRackauckas:
> Without understanding the details, it seems like PowerModels might need https://github.com/lanl-ansi/PowerModels.jl/pull/885

I'll setup a new build that points the manifest at that and see what happens.

PR #722, opened 2023-09-15
Bump OPF size to 10000

ChrisRackauckas:


ChrisRackauckas:
This is failing because of https://github.com/SciML/SciMLBenchmarks.jl/pull/735

PR #769, opened 2023-10-08
[WIP] Fix SimpleHandwrittenPDE wpd

ErikQQY:
See https://docs.sciml.ai/SciMLBenchmarksOutput/dev/SimpleHandwrittenPDE/allen_cahn_fdm_wpd/, many of the simple PDE benchmarks fail. Where should I check the usage of ```LinSolveFactorize```? It seems we no longer use this syntax.

### DIFF
--- a/benchmarks/SimpleHandwrittenPDE/allen_cahn_fdm_wpd.jmd
+++ b/benchmarks/SimpleHandwrittenPDE/allen_cahn_fdm_wpd.jmd
@@ -45,19 +45,13 @@ Reference solution using Vern9 is below:
 
 ```julia
 xs, prob = allen_cahn(100)
-sol = solve(prob, RadauIIA5(autodiff=false); abstol=1e-14, reltol=1e-14, dt=1e-4, adaptive=false)
+sol = solve(prob, Vern9(); abstol=1e-14, reltol=1e-14, dt=1e-4, adaptive=false)
 test_sol = TestSolution(sol);
 
 tslices = [0.0 0.25 0.50 0.75 1.]
 ys = hcat((sol(t) for t in tslices)...)
 labels = ["t = $t" for t in tslices]
 plot(xs, ys, label=labels)
-```
-
-Linear solvers
-
-```julia
-const LS_Dense = LinSolveFactorize(lu)
 ```
 
 ## High tolerances
--- a/benchmarks/SimpleHandwrittenPDE/allen_cahn_spectral_wpd.jmd
+++ b/benchmarks/SimpleHandwrittenPDE/allen_cahn_spectral_wpd.jmd
@@ -37,11 +37,11 @@ function allen_cahn(du,u,x,t)
 end
 ```
 
-Reference solution using RadauIIA5 is below:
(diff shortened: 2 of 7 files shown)


benchmarks/SimpleHandwrittenPDE/allen_cahn_spectral_wpd.jmd
```diff
@@ -37,11 +37,11 @@ function allen_cahn(du,u,x,t)
 end
 ```
 
-Reference solution using RadauIIA5 is below:
+Reference solution using SBDF2 is below:
 
 ```julia
 prob = SplitODEProblem(DiffEqArrayOperator(D2), allen_cahn, w, (0.0,5.0), x)
-sol  = solve(prob, RadauIIA5(autodiff=false); reltol=1e-14,abstol=1e-14)
+sol  = solve(prob, SBDF2(), dt=1e-3; reltol=1e-14,abstol=1e-14)
```
ChrisRackauckas, line 44: we shouldn't use this as a reference solution since it's fixed time step. We need an adaptive algorithm to be reference.

benchmarks/SimpleHandwrittenPDE/allen_cahn_fdm_wpd.jmd
```diff
@@ -45,7 +45,7 @@ Reference solution using Vern9 is below:
 
 ```julia
 xs, prob = allen_cahn(100)
-sol = solve(prob, RadauIIA5(autodiff=false); abstol=1e-14, reltol=1e-14, dt=1e-4, adaptive=false)
+sol = solve(prob, Vern9(); abstol=1e-14, reltol=1e-14, dt=1e-4, adaptive=false)
```
ChrisRackauckas, line 48: Is this equation non-stiff enough for this to work?

ChrisRackauckas:
`LinSolveFactorize` was replaced with `linsolve` and LinearSolve.jl, effectively all the stuff in here: https://docs.sciml.ai/DiffEqDocs/stable/tutorials/advanced_ode_example/

ChrisRackauckas:
Is there a reason why you removed Radau as the reference solution? That's a good method for the refernence.

ErikQQY:
> LinSolveFactorize was replaced with linsolve and LinearSolve.jl, effectively all the stuff in here: https://docs.sciml.ai/DiffEqDocs/stable/tutorials/advanced_ode_example/

I see, I will change to the linear solvers in LinearSolve.jl

> Is there a reason why you removed Radau as the reference solution? That's a good method for the refernence.

Yeah, RadauIIA5 is a great method for stiff equations,  but I tested locally and it always errors with 
```julia
Error: Non-concrete Jacobian not yet supported by RadauIIA5.
```

ChrisRackauckas:
> I see, I will change to the linear solvers in LinearSolve.jl

It can probably just be removed. The default LU is better these days than the one that chooses.

> Yeah, RadauIIA5 is a great method for stiff equations, but I tested locally and it always errors with

Create a separate version of the problem which is not split. We should test non-split methods vs split methods using `prob_choice`.

PR #775, opened 2023-10-17
[WIP] Add Brussletor benchmarks for nonlinear solvers

ErikQQY:
Add Brusselator benchmarks for all the nonlinear solvers

### DIFF
--- a/benchmarks/NonlinearProblem/Bruss_wpd.jmd
+++ b/benchmarks/NonlinearProblem/Bruss_wpd.jmd
@@ -16,7 +16,7 @@ RUS = RadiusUpdateSchemes;
 Define the Brussletor problem.
 
 ```julia
-const N = 32
+const N = 64
 const xyd_brusselator = range(0, stop = 1, length = N)
 brusselator_f(x, y) = (((x - 0.3)^2 + (y - 0.6)^2) <= 0.1^2) * 5.0
 limit(a, N) = a == N + 1 ? 1 : a == 0 ? N : a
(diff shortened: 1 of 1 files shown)


benchmarks/NonlinearProblem/Bruss_wpd.jmd
```diff
@@ -0,0 +1,124 @@
...
+```julia
+using NonlinearSolve, NonlinearSolveMINPACK, SciMLNLSolve, SimpleNonlinearSolve, Sundials
+using LinearAlgebra, SparseArrays, DiffEqDevTools, Plots
+RUS = RadiusUpdateSchemes;
+```
+
+Define the Brussletor problem.
+
+```julia
+const N = 32
```
avik-pal, line 19: Lets add an axis for increasing the discretization.

ErikQQY: Increasing N to increase the discretization? Wouldn't that increases the benchmarking time?

ChrisRackauckas: yes, but we got a big machine

benchmarks/NonlinearProblem/Bruss_wpd.jmd
```diff
@@ -0,0 +1,124 @@
...
+```julia
+using NonlinearSolve, NonlinearSolveMINPACK, SciMLNLSolve, SimpleNonlinearSolve, Sundials
+using LinearAlgebra, SparseArrays, DiffEqDevTools, Plots
+RUS = RadiusUpdateSchemes;
+```
+
+Define the Brussletor problem.
+
+```julia
+const N = 64
```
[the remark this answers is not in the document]
ErikQQY: So we need to increase both N and the order of tolerance to make sure the algorithms scale?

ChrisRackauckas:
We can do this in two steps. Let's first get one in with a fixed N. Then let's do one where we loop over N and show the difference of the methods.

PR #801, opened 2023-12-19
Try merging the projects

avik-pal:


src/SciMLBenchmarks.jl
```diff
@@ -61,17 +69,20 @@ end
...
+      @eval @subprocess begin
+        using SciMLBenchmarks
+        SciMLBenchmarks.weave_file($folder, $file, $build_list)
       end
+    catch e
+      @show folder, file
+      @error(e)
+    end
   end
 end
```
avik-pal, line 87: @ChrisRackauckas What do you think about this? This prevents segfaults at least locally. All the subprocesses are blocking and carry over all the options so they won't interfere with the benchamrks

ChrisRackauckas: That looks fine, just neds a rebase

benchmarks/NonlinearProblem/nonlinear_solver_23_tests.jmd
```diff
@@ -100,10 +100,10 @@ function benchmark_problem!(prob_name; solver_tracker=solver_tracker, selected_N
     solvers_TR = filter(s -> s.type==:TR, successful_solvers)
     solvers_LM = filter(s -> s.type==:LM, successful_solvers)
     solvers_PT = filter(s -> s.type==:PT, successful_solvers)
```
avik-pal, line 102: @yonatanwesen why was this added as a separate solver list? It is a single solver so we don't need a separate plot. Can you open a follow up PR to fix this

avik-pal:
@ChrisRackauckas  this should be good to go

PR #804, opened 2023-12-21
add PT to 23 NonlinearProblem tests

yonatanwesen:
This PR adds PseudoTransient methods to the 23 nonlinear problems tests.


yonatanwesen:
All I had to do was add PT to the solvers list, but it is more involved than I initially expected. I have added all the required changes to ensure plots include PT now.

PR #806, opened 2023-12-22
Fix plots

yonatanwesen:
I have added PT to NR group which should fix the plots now. Can we wait till the plots are generated before merging?


avik-pal:
make it `general`

avik-pal (approved):

PR #809, opened 2023-12-26
Standardize the NonlinearProblem Plots

avik-pal:
Trying to make the plots look all similar and better for the paper

- [x] Battery Problem
- [x] Quadratic Problem
  - [x] Pretty Plots
  - [x] Show more focused plots
- [x] 23 Test Problem
  - [x] Makie
  - [x] Summary Plot
- [ ] ~Bruss~


avik-pal:
I rewrote the plotting for the 23 test problems with makie. thoughts on the new plots

https://julialang-buildkite-artifacts.s3.amazonaws.com/scimlbenchmarks/6af2a6a5dc978fbf636526edf028c4067aea022bbb0392bcc00f8b76a54c0bbe/markdown/NonlinearProblem/trigonometric_function_wpd.svg

TorkelE:
Looks good, although I am not sure what the advantage is more than not using Plots (which I prefer). There colors in the example (for the red and the dark one) is hard to distinguish). Previously I tried to use different colors to mark the different categories, I thought that looked quite good:
image

avik-pal:
Makie.jl is a bit more (easily) configurable (once you get used to it) and their docs are much nicer IMO (especially from REPL). Re the colors, I tend to prefer a bit more sober colors for publications, but the color scheme can be changed quite easily

TorkelE:
Cannot really argue with Makie seeming more potent, just haven't taken to liking their set up.

Generally, for colors, I think that using a wider range, and making things almost "cartoon-like" (thick text borders etc.) makes things much easier to read.

avik-pal:
@ChrisRackauckas feel free to merge. I will work on the bruss in a separate PR to prevent triggering the build for these 3 benchmarks repeatedly.

Once https://github.com/SciML/SimpleNonlinearSolve.jl/pull/125 is merged we can bring back the Manifest to main

PR #832, opened 2024-01-16
Testing the new NonlinearSolve.jl Performance

avik-pal:


avik-pal:
Actually let's not merge this. Once https://github.com/SciML/NonlinearSolve.jl/pull/345 is in, I will tag a release and update the manifest for this

PR #836, opened 2024-01-21
Add OptimizationFrameworks/clnlbeam.jmd

odow:
This is a model similar to the OPF example, except that it is much simpler and has no type stability issues. This makes it a simpler candidate for benchmarking during development.

It is taken from the JuMP docs: https://jump.dev/JuMP.jl/stable/tutorials/nonlinear/simple_examples/#The-clnlbeam-problem
As a goal, we're looking to solve the N = 1000 case, although the `jmd` file only goes up to `N = 200`.

Am I meant to update the `Manifest.toml`? How does that work.

I get a graph like this, but I was doing stuff on my machine while it executed (and it's really hot here, so the fan was going full bore), so results may differ. At the very least, it's a good candidate model for testing the structural simplification of MTK, and it teases apart different differentiation implementations.

image

## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
- [x] Any new documentation only uses public API


odow:
That seemed to work. Here's the CI plot:

ChrisRackauckas:
Awesome, thanks! While you got this running, can you also share a quick StatProfilerHTML on the AutoSparseReverseDiff(true) one on one of the bigger sizes to see if that confirms overhead dominated by sparsity detection?

ChrisRackauckas:
Awesome, yup that's the benchmark we've been looking for.

PR #839, opened 2024-01-24
[WIP] Fix 23 Test Problems Benchmarking

avik-pal:
Testing the cost of the cache construction. `linsolve = nothing` triggers a high cost especially for small problems due the Krylov solvers. If this works well, we should consider a type stable automatic determination of the linear solver in NonlinearSolve if no linear solver is specified.


ChrisRackauckas:
I think we can call in our Krylov.jl buddies and get the root issue fixed here. The cache construction that you're talking about is a specialized zero initialization already https://github.com/SciML/LinearSolve.jl/blob/main/src/iterative_wrappers.jl#L176-L189 If Krylov.jl is taking a ton of time to construct a type-matching cache from that, then there's something going wrong and it's not specializing on the empty case effectively. @amontoison can you look into what the main contributor of the cache construction time is for that and if it can be specialized to empty arrays / matrices?

avik-pal:
- [ ] Specify chunksize (type stability)
- [x] Specify Linear Solver
- [ ] Upstream Patch Merge
- [ ] Add Simple Solvers
- [ ] Fix the plotting -- new termination codes are not handled.

ChrisRackauckas:
Note that we probably need to setup specialization choices on NonlinearSolve.jl for compile times anyways, and that would fix type stability chunk sizes by default.

amontoison:
> I think we can call in our Krylov.jl buddies and get the root issue fixed here. The cache construction that you're talking about is a specialized zero initialization already https://github.com/SciML/LinearSolve.jl/blob/main/src/iterative_wrappers.jl#L176-L189 If Krylov.jl is taking a ton of time to construct a type-matching cache from that, then there's something going wrong and it's not specializing on the empty case effectively. @amontoison can you look into what the main contributor of the cache construction time is for that and if it can be specialized to empty arrays / matrices?

Hi @ChrisRackauckas and @avik-pal
I can have a look next week if that's fine for you.

ChrisRackauckas:
Thanks, a week is totally fine.

amontoison:
@avik-pal @ChrisRackauckas 
Sorry for not answering ealier.
I checked you [file](https://github.com/SciML/LinearSolve.jl/blob/main/src/iterative_wrappers.jl#L176-L189) and I don't understand why you want to initialize with an empty matrix.
The workspace of a Krylov method (`KrylovSolver`) requires the size of the linear system that you want to solve like `CgSolver(m, n, S)` with `S=Vector{Float64}` (or anything similar for GPU arrays).
You can call `CgSolver(A, b)` but it will just do `m, n = size(A)` and `S=ktypeof(b)`.

amontoison:
Off-topic: For this function: https://github.com/SciML/LinearSolve.jl/blob/main/src/iterative_wrappers.jl#L97
, I defined a dictionary in Krylov.jl:
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/krylov_solvers.jl#L13

ChrisRackauckas:
> I checked you [file](https://github.com/SciML/LinearSolve.jl/blob/main/src/iterative_wrappers.jl#L176-L189) and I don't understand why you want to initialize with an empty matrix.

We are initiating a sum type, so there's a very high probability that the workspace won't be used at all. But an object of the correct type is still required.

PR #857, opened 2024-02-10
Brusselator Benchmarks [Take III]

avik-pal:
## Checklist

- [ ] Work Precision Diagram for a fixed N
- [ ] Scaling
  - [x] First Order Solvers
  - [ ] Jacobian Free Methods


avik-pal:
@ChrisRackauckas I think we can merge this. I will work on the 2 remaining plots in a separate PR to prevent retriggering all the benchmarks

PR #859, opened 2024-02-10
Rerun the LU Factorization Benchmarks

avik-pal:


docs/pages.jl
```diff
@@ -40,12 +40,13 @@ end
...
 
 permute!(pages,
-    [1, 13, 11, 9, 19, 17, 27, 4, 6, 15, 25, 5, 7, 14, 18, 28, 16, 26, 10, 8, 21, 20, 24, 3, 12, 22, 23, 2, 29]
+    [1, 14, 11, 12, 9, 20, 18, 28, 4, 6, 16, 26, 5, 7, 15, 19, 29, 17, 27, 10, 9, 22, 21, 25, 3, 13, 23, 24, 2, 30]
 )
 
 names = [
     "SciMLBenchmarks.jl: Benchmarks for Scientific Machine Learning (SciML) and Equation Solvers",
     "Multi-Language Wrapper Benchmarks",
+    "LU Factorization",
```
ChrisRackauckas, line 49: The other ones are also LU factorization

avik-pal: I just wanted to separate out Pardiso which is holding back MKL_jll

ChrisRackauckas: Is the new JLL any better?

ChrisRackauckas: If you want to split it, it would be dense vs sparse

avik-pal: The benchmarks seems consistent with the older version. The benchmarks I saw could just be a artifact of Ryzen vs EPYC

PR #880, opened 2024-02-24
Add COLNEW benchmark

ErikQQY:
Add benchmarks for COLNEW method


benchmarks/NonlinearProblem/nonlinear_solver_23_tests.jmd
```diff
@@ -565,7 +565,6 @@ end
 save("summary_wp_23test_problems.svg", fig)
 ```
 
-# Appendix
```
ChrisRackauckas, line 568: Revert this? I don't understand why we would remove the standard appendix header

ErikQQY: This one is duplicated, see:
https://github.com/SciML/SciMLBenchmarks.jl/blob/edcbe452deb980f9379204344bc6e525067cd846/src/SciMLBenchmarks.jl#L89-L96

And

https://docs.sciml.ai/SciMLBenchmarksOutput/dev/NonlinearProblem/nonlinear_solver_23_tests/#Appendix

ChrisRackauckas: gotcha

ChrisRackauckas:
This should get rerun

PR #892, opened 2024-04-08
Add a section for automatic differentiation benchmarks

ChrisRackauckas:
DifferentiationInterface doesn't match standard Pkg conventions so it's having issues installing, but this should get the infra together.


benchmarks/AutomaticDifferentiation/JuliaAD.jmd
```diff
@@ -0,0 +1,33 @@
+---
+title: Julia AD Benchmarks
+author: Chris Rackauckas
```
gdalle, line 3: Strictly speaking I'm the author of that code ;)

ChrisRackauckas: Feel free to edit it in a PR

PR #893, opened 2024-04-08
Fix DifferentiationInterface loading in AD benchmarks

gdalle:
Following the instructions to install from a subdirectory of a Git repo:

```julia
julia> using Pkg

julia> Pkg.add(
        url="https://github.com/gdalle/DifferentiationInterface.jl",
        subdir="DifferentiationInterface"
    )
    
julia> Pkg.add(
        url="https://github.com/gdalle/DifferentiationInterface.jl",
        subdir="DifferentiationInterfaceTest"
    )
```

Not sure why that failed on Chris' computer


gdalle:
@ChrisRackauckas I saw that your `Manifest.toml` was generated with Julia v1.10.0, mine was with the latest v1.10.2. Do you think the upgrade could solve your installation problems?

gdalle:
Something I don't like about the display of the dataframe is that the floating point columns are not easily comparable. It would be easier if they were all in scientific notation like `1.93e-3` but I don't know how to achieve that

ChrisRackauckas:
Ehh it can always improve. This is at least a good v1. I think other ADs and such are probably next on the list.

ChrisRackauckas:
> @ChrisRackauckas I saw that your Manifest.toml was generated with Julia v1.10.0, mine was with the latest v1.10.2. Do you think the upgrade could solve your installation problems?

That could be it. Or Windows.

gdalle:
thanks for the merge! excited to work with you on this :)

PR #919, opened 2024-04-13
Create cantilever.jmd

Spinachboul:
Add a new surrogates benchmark!

PR #920, opened 2024-04-13
Update Manifest.toml

Spinachboul:


ChrisRackauckas:
This, plus a change to the Project.toml, and https://github.com/SciML/SciMLBenchmarks.jl/pull/919 should all be in the same PR.

Spinachboul:
Oh ok, I'll converge them into same PR!

PR #926, opened 2024-04-18
Added MatrixDepot.jmd file to run benchmarks on sparse matrices

anastasia21112:
## Running Benchmarks
Run benchmarks on a single sparse matrix from the matrix depot suite sparse collection


ChrisRackauckas:
Missing the package in the manifest/project

PR #969, opened 2024-06-17
fix: add more dymola times, use a tolerance of 1e-6

avinashresearch1:
Add more timings for RC circuit with Dymola for different values of `N`. Use tolerance of `1e-6` for simulation tolerance.


avinashresearch1:
@ChrisRackauckas FYI, this PR has updated dymola times.

PR #970, opened 2024-06-18
remove ODE.jl, update solver versions

oscardssmith:


benchmarks/NonStiffODE/LotkaVolterra_wpd.jmd
```diff
@@ -9,7 +9,7 @@ author: Chris Rackauckas
 The purpose of this problem is to test the performance on easy problems. Since it's periodic, the error is naturally low, and so most of the difference will come down to startup times and, when measuring the interpolations, the algorithm choices.
 
 ```julia
-using OrdinaryDiffEq, ParameterizedFunctions, ODE, ODEInterfaceDiffEq, LSODA,
```
ChrisRackauckas, line 12: won't you need to remove the solvers that use it too?

oscardssmith: I believe all the solvers we are using from it are in ODEInterfaceDiffEq (but I'll double check).

oscardssmith:
it looks like the MTK v9 upgrade to the stiff case doesn't work automatically...

oscardssmith:
This seems to have worked now!

benchmarks/StiffODE/VanDerPol.jmd
```diff
@@ -139,7 +139,7 @@ setups = [Dict(:alg=>Rosenbrock23()),
           Dict(:alg=>Rosenbrock23(), :prob_choice => 2),
           Dict(:alg=>Rodas3()),
           Dict(:alg=>TRBDF2()),
-          Dict(:alg=>rodas()),
```
ChrisRackauckas, line 142: why? We want to benchmark against Hairer.

benchmarks/StiffODE/VanDerPol.jmd
```diff
@@ -320,7 +319,6 @@ setups = [Dict(:alg=>Rodas3()),
           Dict(:alg=>CVODE_BDF()),
           Dict(:alg=>Rodas4()),
           Dict(:alg=>Rodas4(), :prob_choice => 2),
-          Dict(:alg=>rodas()),
```
ChrisRackauckas, line 323: why remove Hairer?

ChrisRackauckas:
Rosenbrock static took a major performance hit?

benchmarks/StiffODE/VanDerPol.jmd
oscardssmith: it was hanging

ChrisRackauckas: Add a note about that. There's always a notes on why some methods weren't benchmarked.

oscardssmith:
How are you looking at the results?

oscardssmith:
To me it looks like the static rosenbrocks are faster?

ChrisRackauckas:
See Hires

ChrisRackauckas:
But add the note so we can merge.

oscardssmith:
note added.

PR #978, opened 2024-06-20
MTK Benchmarks

ChrisRackauckas:

PR #981, opened 2024-06-22
ThermalFluid script fix

chriselrod:
Hopefully addressing the concern raised in #978


chriselrod:
thermalfluid_mtk_vs_jsir
thermalfluid_totaltime

ChrisRackauckas:
I'll merge because the total time parts are done, but two of the figures had an issue.

```
Error: Number of elements not equal: 8 content elements and 7 labels.
```

```
Error: Number of elements not equal: 2 content elements and 3 labels.
```

@chriselrod can you follow up with a fix to those?

PR #982, opened 2024-06-24
A few improvements to MatrixDepot benchmarks

ChrisRackauckas:
1. Make the try/catch be per matrix, so that it can continue onto the next matrix. This is preparing it for a larger run.
2. Setup a mean factorization time display. In theory this should be the mean over classes of matrices, but for now this is just the mean over all.
3. Removes the extra functions, keep it simple

### DIFF
--- a/benchmarks/LinearSolve/MatrixDepot.jmd
+++ b/benchmarks/LinearSolve/MatrixDepot.jmd
@@ -53,7 +53,6 @@ for z in 1:length(matrices)
         end
         p = bar(algnames, times[z, :];
             ylabel = "Time/s",
-            labels = algnames,
             yscale = :log10,
             title = "Time on $(matrices[z])",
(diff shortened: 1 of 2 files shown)


benchmarks/LinearSolve/MatrixDepot.jmd
```diff
@@ -21,82 +21,57 @@ algs = [
...
-
-```julia
-run_and_plot(2)
+for z in 1:length(matrices)
+    try
+        rng = MersenneTwister(123)
+        A = mdopen(matrices[z]).A
+        A = convert(SparseMatrixCSC, A)
+        n = size(A, 1)
+        @info "dim=$(dim): $n × $n"
```
ChrisRackauckas, line 42:
```suggestion
        @info "$n × $n"
```

PR #985, opened 2024-06-25
Further fixes for MTK Benchmarks

chriselrod:
Some of the sizes were wrong.


chriselrod:
thermalfluid_julia
thermalfluid_totaltime

chriselrod:
thermalfluid_totaltime

ChrisRackauckas:
That's a CI issue I pointed out to @thazhemadam . Just bump any bound in the Project.toml so that it's not file-specific but folder-specific and the CI is fixed. We should fix it but it's easy to work around.

ChrisRackauckas:
ThermalFluid looks done

RCCircuit still has

```
Error: Number of elements not equal: 9 content elements and 8 labels.
```

PR #1001, opened 2024-06-28
Summary plots on Brusselator sparse AD benchmarks

gdalle:
Run Brusselator benchmarks over a range of input sizes and plot the comparative results.

detection
coloring
differentiation
summary


ChrisRackauckas:
The one I really want to see is FastDifferentiation.jl in this as well.

PR #1007, opened 2024-07-03
Transistor Amplifier Benchmark added.

maxesit:
I have tried to implement the problem the same way Checmical Akzo Nobel is implemented.

## Notes: the code returns a few problems
- [ ] `prob = ODEProblem(sys, u0, tspan)` asks for structural simplify to be run, despite sys being complete | solved by using ss instead of sys
- [x] `CVODE_BDF()` says it is unable to use mass matrices
- [ ] setting solve tolerances to 1/10^14 returns retcode errors

Please let me know if anything else needs to be done( regarding documentation about the benchmark, etc.)

-Polic Marko


benchmarks/DAE/TransistorAmplifier.jmd
```diff
@@ -0,0 +1,266 @@
...
+odaeprob = ODAEProblem(ss, u0, tspan)
+ode_ref_sol = solve(odaeprob, CVODE_BDF(), abstol = 1/10^14, reltol = 1/10^14)
+# CVODE_BDF -> unable to use mass matrices
+
+
+du = mmprob.f(mmprob.u0, mmprob.p, 0.0)
+du0 = D.(unknowns(ss)) .=> du
+daeprob = DAEProblem(ss, du0, [], tspan)
+ref_sol = solve(daeprob, IDA())
+#, abstol = 1/10^14, reltol = 1/10^14) -> retcode: Initial failure
```
ChrisRackauckas, line 110:
```suggestion
du = mmprob.f(mmprob.u0, mmprob.p, 0.0)
du0 = D.(unknowns(ss)) .=> du
daeprob = DAEProblem(ss, du0, [], tspan)
ref_sol = solve(daeprob, IDA())
#, abstol = 1/10^14, reltol = 1/10^14) -> retcode: Initial failure
```

Can you write the DAEProblem directly and see if it's an issue?

maxesit: Retcode issues only start appearing with tolerances tighter than 1/10^8. Is that enough, or do you still want me to try the manual way?

ChrisRackauckas: Can we use a different solver to get a tigher reference solution then? 1e-8 isn't that great of a reference given some of the benchmarks go lower.

maxesit: I've managed to find a solver to raise it to 1e-9, but I guess that isn't much better either?
* DImplicitEuler -> retcode: MaxItters for anything better than 1e-6
* IDA                  -> retcode: Unstable for anything better than 1e-8
* DFBDF             -> retcode: Unstable for anything better than 1e-9 !!!
* DABDF2           -> retcode: Unstable
* DASKR             -> MethodError

ChrisRackauckas: What about as a directly written mass matrix ODE on Rodas5P?

maxesit: using Rodas5P allowed me to increase ODEProblem tolerances to:
`sol = solve(mmprob, Rodas5P(), abstol = 1e-14, reltol = 1e-10)`

that still leaves DAEProblem with 1e-9 tolerances at best:
`ref_sol = solve(daeprob, DFBDF(), abstol = 1e-9, reltol = 1e-9)`

( I've tried adding `mass_matrix = dirMassMatrix` but it doesn't seem to change anything. Should I try implementing the DAEProblem manually and see if that helps w anything?)

ChrisRackauckas: yeah that could just be difficulties due to the properties of fully implicit BDFs with respect to factorization error. Let's use Rodas5P with that as the reference solution then, and then in the other plots see how well the others converge.

maxesit:
I've added the [Car Axis benchmark](https://archimede.dm.uniba.it/~testset/report/caraxis.pdf).
In the problem formulation file, there're $\epsilon$ and $\varepsilon$. However, the first one isn't used in the formulas, and the latter doesn't have a prescribed value. I assume this is a typo, though it may not be!

I've checked for different solver algs, but they return retcode: Unstable.

ps. Due to the exams this week, I won't be able to dedicate time until Saturday, but if something more needs to be fixed, I'll be more than happy to do so then.

ChrisRackauckas:
Can you start a new branch for the new one? This was just about to merge (needs to investigate something with MTK, but it seems just about ready to go)

maxesit:
Of course! Will do it in 1 hour.

PR #1012, opened 2024-07-18
Bump stiff ODE benchmarks

ChrisRackauckas:


ChrisRackauckas:
@oscardssmith it looks like something is up with HIRES static arrays, can you look into that ASAP?

PR #1022, opened 2024-07-28
Bump Bio for Catalyst v14 and MTK v9

ChrisRackauckas:


ChrisRackauckas:
@TorkelE @isaacsas this is the one to watch for the major FBDF changes

isaacsas:
I’d imagine many of these need updating for Catalyst 14 (eg ReactionNetworkImorters generated systems need to be marked complete).

ChrisRackauckas:
@isaacsas can you help with what needs to be changed with this bump?

isaacsas:
@ChrisRackauckas LMK if this doesn't work. I think all that was needed was to complete the converted ODESystem, and get the `u0` and `p` from the `oprob` now (since `varmap_to_vars` no longer gives the right format for `p`). 

On a sidenote, it would be useful if the parameter to MTKParameter conversion functionality could be factored into an exported user-accessible function so users can manually process parameter maps to the new form. (Essentially a `varmap_to_vars` replacement.)

ChrisRackauckas:
> On a sidenote, it would be useful if the parameter to MTKParameter conversion functionality could be factored into an exported user-accessible function so users can manually process parameter maps to the new form. (Essentially a varmap_to_vars replacement.)

Interesting point. Open an issue.

avik-pal:
How long is this benchmark? It seems to error https://buildkite.com/julialang/scimlbenchmarks-dot-jl/builds/2705#019146bf-645a-4b43-b1db-597856afaf3b/403-614 but is holding up the CI for more than a day :sweat:

isaacsas:
Maybe @TorkelE knows how long it should take. It is a long one but I have no idea if it should be days.

ChrisRackauckas:
Well it's done now. Looks like it takes 60 hours.

TorkelE:
60 hours doesn't seem too outlandish for the BCR one.

isaacsas:
Has the BCR sparse Jacobian compiliation gotten worse? 80 minutes now seems pretty bad and worse than I remember.

Also @ChrisRackauckas seems like there are Symbolics related errors in the WP plotting now?

ChrisRackauckas:
I don't think it's symbolic, but yes I need to investigate what happened there.

isaacsas:
@ChrisRackauckas is this good to merge now? @vyudu has another PR to add a new benchmark but we should wait in that one till the updated Project.toml and Manifest.toml are merged.

ChrisRackauckas:
This still has a few failures IIUC from the CI

ChrisRackauckas:
Okay I'm going to merge and just fix Bikhori separately since that requires a lot less compute.

PR #1024, opened 2024-08-01
Allow some more matrices

ChrisRackauckas:


ChrisRackauckas:
@staticfloat is there a way to tell if we are ooming the machine, and maybe delete some of the downloaded files on the fly?

staticfloat:
I looked through the `dmesg` logs and there are no OOM or segfault errors.  Looking at the `Timeline` view on the buildkite UI, it looks like it just reached maximum execution time:

image

ChrisRackauckas:
I don't understand what's stalling then. It's just downloading more matrices and running them, but it doesn't fail in those steps, it fails just when writing something to a file?

ChrisRackauckas:
The plot is made to be a png so its size is constant?

staticfloat:
How long is this job supposed to take? This times out after two days.

ChrisRackauckas:
I just don't understand why adding a few more matrices lets the analysis run, but then the weave just stops building? I would've thought the actual work was the hard part, but it seems to get stuck on something non-work when I bump the number of matrices up.

ChrisRackauckas:
I'll bump up the maximum time and see if that works.

ChrisRackauckas:
@anastasia21112 it looks like the mean factorization time graph is not robust to having some drop out. That's worth fixing.

I'll do another size bump and see what that takes. If it's like a week then we're fine.

PR #1026, opened 2024-08-11
[WIP] perf: common neural network workloads benchmarks

avik-pal:
currently add CPU versions. should be easy to extend to GPU once SciMLBenchmarks has GPU runners available.


ChrisRackauckas:
@thazhemadam @staticfloat what's the easiest way to set this up so that benchmarks can choose a separate runner for GPUs?

avik-pal:
Allowing different runners based on the jmd metadata. Something like

```md
---
title: Simple Neural Networks
author: Avik Pal
backend: CUDA <--- If nothing present then use CPU. In future we can allow AMDGPU/Metal/etc with the same syntax
---
```

avik-pal:
Lux now matches SimpleChains in inference timings :sweat_smile:. The ones where we fall behind are because Octavian is somewhat slow on EPYC machines so it is turned off

PR #1046, opened 2024-09-10
Create astrochem benchmark

ChrisRackauckas:


ChrisRackauckas:
```
ERROR: LoadError: InitError: could not load library "/cache/julia-buildkite-plugin/depots/5b300254-1738-4989-ae0a-f4d2d937f953/artifacts/cb7fc2801ca0133a5bdea4bc4585d07c08284cfa/lib/libsundials_sunlinsolklu.so"
```

KLU is having build failures. @staticfloat is there something that might've changed in the SuiteSparse builds?

staticfloat:
Not that I'm aware of; those should be pretty much locked down to the julia version under user in any case.

benchmarks/AstroChem/nelson.jmd
```diff
@@ -0,0 +1,173 @@
+---
+title: Nelson Work-Precision Diagrams
+author: Stella Offner and Chris Rackauckas
+---
+
+```julia
+using OrdinaryDiffEq
+using Plots
```
ChrisRackauckas, line 8:
```suggestion
using DiffEqDevTools, Plots
```

PR #1066, opened 2024-09-23
Plotting Spaced out Sparsity

anastasia21112:
## Changes
- Reduce every matrix to 32x32 
- For every 2x2 block, count how many zeros and if >= 3, replace block with 1, else 0 
- Multiply the amount of nonzero elements in the 32x32 matrix by percentage sparsity
- Plot this metric against the time for every algorithm


ChrisRackauckas:
The plot seems to have failed.

PR #1067, opened 2024-09-25
DynamicalODE Changes

ParamThakkar123:
## Checklist

- [X] Appropriate tests were added
- [] All documentation related to code changes were updated
- [X] The new code follows the
- [X] Any new documentation only uses public API


ChrisRackauckas:
No that's backwards in compat.

ParamThakkar123:
@ChrisRackauckas any changes needed further ??

ChrisRackauckas:
`Error: UndefVarError: `singlependulum` not defined`.

ChrisRackauckas:
`Error: UndefVarError: `plotsolenergy` not defined`

ParamThakkar123:
> `Error: UndefVarError: `singlependulum` not defined`.

I think this is already present here : 

https://github.com/SciML/SciMLBenchmarks.jl/blob/master/benchmarks/DynamicalODE/single_pendulums.jmd

Where can I make an update if this needs a fix ??

ChrisRackauckas:
The point of the project is to get that benchmark set working on the latest versions 😅 . That's in the benchmark script of the folder.

ParamThakkar123:
Oh, Got it 😅

ParamThakkar123:
But my question was why did we get a undef error if that's already defined there ??

ChrisRackauckas:
Previous bad bump maybe?

ParamThakkar123:
@ChrisRackauckas How can I see the errors that have occurred ?? 

> `Error: UndefVarError: `plotsolenergy` not defined`

Like here

ErikQQY:
@ParamThakkar123 Check the README of this repo: https://github.com/SciML/SciMLBenchmarks.jl?tab=readme-ov-file#inspecting-benchmark-results, and the built benchmarks can be checked in the corresponding md files

ErikQQY:
Seems the errors are caused by PyPlot.jl

You may consider figuring out what's wrong with PyPlot.jl here or change to Plots.jl(recommend)

ParamThakkar123:
> Seems the errors are caused by PyPlot.jl
> 
> You may consider figuring out what's wrong with PyPlot.jl here or change to Plots.jl(recommend)

The errors says that the figure function is not defined. Though it's there in PyPlot

ErikQQY:
You can try to update the compat and the Manifest of the packages in this benchmark to the latest version, many of them are outdated

ChrisRackauckas:
It would probably be best to just update the plots to use Plots.jl

ParamThakkar123:
> It would probably be best to just update the plots to use Plots.jl

Noted. I will make the changes

benchmarks/DynamicalODE/Manifest.toml
```diff
@@ -2345,4 +2345,4 @@ version = "3.5.0+0"
 deps = ["Artifacts", "JLLWrappers", "Libdl", "Pkg", "Wayland_jll", "Wayland_protocols_jll", "Xorg_libxcb_jll", "Xorg_xkeyboard_config_jll"]
 git-tree-sha1 = "9c304562909ab2bab0262639bd4f444d7bc2be37"
 uuid = "d8fb68d0-12a3-5cfd-a85a-d49703b185fd"
-version = "1.4.1+1"
+version = "1.4.1+1"
```
ChrisRackauckas, line 2348: this manifest bump is not correct!

ChrisRackauckas:
The updates to the plots are correct, but the manifest was not bumped so this isn't actually testing the newer versions. Can you bump the manifest? Otherwise we need to revert.

ParamThakkar123:
Yes @ChrisRackauckas , I will create a new pr by today which bumps the manifest with the latest versions for both DynamicalODE and BayesianInference.

PR #1074, opened 2024-09-28
Update sparse AD benchmarks

gdalle:
Update sparse AD benchmarks to latest versions of DifferentiationInterface, SparseMatrixColorings and SparseConnectivityTracer


gdalle:
@ChrisRackauckas any idea why the website still hasn't updated? Am I looking at the wrong URL?

ChrisRackauckas:
I had to manually unblock a step.

PR #1077, opened 2024-10-02
perf: testing the newer nonlinear solve versions

avik-pal:
mostly to check https://github.com/SciML/NonlinearSolve.jl/pull/468

I am nearly certain I need to make other changes to get this to work but lets see


avik-pal:
there was probably a merge with the signature not updated. @thazhemadam can you update it on master?

thazhemadam:
Done in https://github.com/SciML/SciMLBenchmarks.jl/pull/1078.

ChrisRackauckas:
@avik-pal it looks like it worked? Let me know if this should revert, but it looks fine to me.

avik-pal:
This is using a commit which won't exist now. Let's revert for now. I want to refresh the benchmarks next week with DI's sparsity stuff

PR #1079, opened 2024-10-05
Update linear BVP benchmarking

ErikQQY:
Update all the linear BVP benchmarking scripts and figures, and switch to CairoMakie for a better diagram


ErikQQY:
Seems the benchmarks need to be unblocked?

ChrisRackauckas:
The FIRK methods seem to have issues in their adaptivity algorithm? But looks like the benchmarks worked!

ErikQQY:
> The FIRK methods seem to have issues in their adaptivity algorithm?

Yeah, the adaptivity in FIRK methods is weird, I need to investigate the causes.

Also, the COLNEW wrapper seems to fail on some problems, potentially there are some bugs in the COLNEW wrapper.

PR #1097, opened 2024-10-15
Remove unsuccessful solvings in linear BVP benchmarks

ErikQQY:
`COLNEW` in linear BVP benchmarks failed on some problems, this PR adds successful solving checks


ChrisRackauckas:
Make the change in DiffEqDevTools to reject non successful retcodes

PR #1102, opened 2024-10-21
Update manifest.toml

ParamThakkar123:
## Checklist

- [X] Appropriate tests were added
- [X] Any code changes were done in a way that does not break public API
- [X] The new code follows the
- [X] Any new documentation only uses public API

### DIFF
--- a/benchmarks/DynamicalODE/Manifest.toml
+++ b/benchmarks/DynamicalODE/Manifest.toml
@@ -1,19 +1,18 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.10.4"
+julia_version = "1.11.0"
 manifest_format = "2.0"
-project_hash = "fa5155789e53e26235ac09f5e9324aff13202249"
+project_hash = "e9c32030429ed270bc56c8290ff46760e0e9f094"
 
 [[deps.ADTypes]]
-git-tree-sha1 = "016833eb52ba2d6bea9fcb50ca295980e728ee24"
+git-tree-sha1 = "eea5d80188827b35333801ef97a40c2ed653b081"
 uuid = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
-version = "0.2.7"
+version = "1.9.0"
+weakdeps = ["ChainRulesCore", "EnzymeCore"]
 
-[[deps.AbstractAlgebra]]
-deps = ["GroupsCore", "InteractiveUtils", "LinearAlgebra", "MacroTools", "Preferences", "Random", "RandomExtensions", "SparseArrays", "Test"]
-git-tree-sha1 = "d7832de8cf7af26abac741f10372080ac6cb73df"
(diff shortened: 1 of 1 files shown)


ParamThakkar123:
@ErikQQY getting this error in `Henon-Heiles_energy_conservation_benchmark.md` . Doesn't look like it's related to the versions

ErikQQY:
I am sure this is caused by the version bumps, the "Quadruple Boson Energy Conservation" also failed because of this error, you may need to see what breaking changes led to this.

ParamThakkar123:
@ErikQQY the error occurred because in DiffEqCallbacks.jl the autodiff value is set to nothing. And these changes in DiffEqCallbacks are causing this issue.

image

ErikQQY:
The errors might caused by the outdated syntax, you can refer to the latest docs and see how to use the right syntax to fix the errors.

benchmarks/DynamicalODE/Henon-Heiles_energy_conservation_benchmark.jmd
```diff
@@ -18,6 +18,8 @@ In order to separate these two, we will use `iip` for the in-place names and `oo
 ```julia
 using OrdinaryDiffEq, Plots, DiffEqCallbacks
 using SciMLBenchmarks
+using ADTypes
+using DifferentialEquations
```
ChrisRackauckas, line 22: you don't need this

ParamThakkar123: Made the changes. Actually I searched some docs. So it pointed to these dependencies

ParamThakkar123:
@ChrisRackauckas the error shows that an autodiff or manifold_project should be used here. Which one should I use ?? AutoForwardDiff ??

ParamThakkar123:
> @ChrisRackauckas the error shows that an autodiff or manifold_project should be used here. Which one should I use ?? The ForwardDiff from ForwardDiff module ??

@ErikQQY can you please help me with this part ? I tried everything but a new error occurs always, and there's cyclic dependency

ParamThakkar123:
@ErikQQY According to the docs `manifold_jacobian` is a parameter of the `ManifoldProjection` function. But while using it, it gives a method error saying there's nothing like that. Without using `manifold_jacobian` everything works fine locally but when I push the code the I get a method error in the CI

ParamThakkar123:
The latest syntax mentioned in the docs doesn't seem to work at all

ErikQQY:
>The latest syntax mentioned in the docs doesn't seem to work at all

You might be doing something wrong with the syntax

Here are some useful tips you may find useful:
1. Since this benchmark is not "that big" you can debug in your local environment and see what's going wrong, instead of waiting for CI to finish.
2. Ensure your compats and Manifest are up-to-date.
3. If you have trouble understanding how to correctly use the package, docs and test files are helpful.

PR #1116, opened 2024-11-04
feat: update to nonlinear solve v4

avik-pal:


avik-pal:
I want to get the PETSc ones in as well but those keep segfaulting on my local computer. So ig we could move forward with this and I follow up with that soon

PR #1137, opened 2024-12-05
feat: address reviewer comments about the benchmarks

avik-pal:


ChrisRackauckas:
Is this stale? I think so right?

avik-pal:
yeah all of them got added via the other PR

PR #1138, opened 2024-12-16
Add new benchmark for sparse AD on simple loop written diffusion operator

yolhan83:
This PR adds a new benchmark to test sparse automatic differentiation (AD) methods on a simple diffusion operator implemented using a loop. It evaluates various AD backends, including Enzyme, ForwardDiff, Mooncake, and sparse configurations with `TracerSparsityDetector` and `GreedyColoringAlgorithm`. A manual Jacobian is provided for validation, and the benchmark framework compares performance and accuracy for a random input vector of size 1000.


ChrisRackauckas:
```
Error: ArgumentError: Package DifferentiationInterfaceTest not found in cur
rent path.
- Run `import Pkg; Pkg.add("DifferentiationInterfaceTest")` to install the 
DifferentiationInterfaceTest package.
```

The project/manifest need to get updated.

yolhan83:
It should work now, thank you

yolhan83:
wait were the tests suppose to be in 1.10.5 ?

ChrisRackauckas:
I'm not sure if the table is nicely displayed? But we'll see in the dev version of the docs. The results look nice and useful though.

yolhan83:
If it's not there is the parse from the Julia AD Benchmark which can be used, will see

PR #1139, opened 2024-12-20
correct parsing of the DF

yolhan83:
Correct the parsing of the dataframe in the benchmark Diffusion operator loop sparse AD benchmarks


ChrisRackauckas (approved):

ChrisRackauckas:
Perfect thanks!

PR #1144, opened 2024-12-27
MatrixIdentification.jl

anastasia21112:
Create MatrixIdentification.jl to classify any sparse matrix into a different structured matrix (Banded, Symmetric, Hermitian, etc.) and getstructure function to return percentage sparsity or bandedness.


ChrisRackauckas:
Make this into a package.

PR #1164, opened 2025-01-23
Add Symbolics hashconsing benchmarks

AayushSabharwal:


AayushSabharwal:
Only the Symbolics/ThermalFluid.jmd benchmark is relevant. I'll remove the other one

PR #1178, opened 2025-02-20
feat: add benchmark for effect of CSE on jacobian computation/compilation

AayushSabharwal:


ChrisRackauckas:
What's the state of this?

benchmarks/Symbolics/old_sparse_jacobian.jl
```diff
@@ -0,0 +1,227 @@
+function old_sparsejacobian(ops::AbstractVector, vars::AbstractVector)
+    sp = Symbolics.jacobian_sparsity(ops, vars)
```
ChrisRackauckas, line 2: Should be updated to DI?

AayushSabharwal: This is a recreation of the old Symbolics AD algorithm to compare before and after. An apples-to-apples comparison would use `jacobian_sparsity` here as well, since that is what Symbolics uses.

PR #1230, opened 2025-05-19
Benchmarking Variable Rate Aggregator

sivasathyaseeelan:
## Checklist

- [ ] Appropriate tests were added
- [ ] The new code follows the
- [ ] Any new documentation only uses public API


ChrisRackauckas:
Needs to rebase to run

ChrisRackauckas:
Rebase, you have other changes in this PR.

ChrisRackauckas:
The manifest needs a bump, and it needs to be run on v1.10

sivasathyaseeelan:
> The manifest needs a bump, and it needs to be run on v1.10

sivasathyaseeelan:
> Were you going to move this to the right folder?

Yes I will move the tests

ChrisRackauckas:
@isaacsas it looks like FRM is faster until about 5-10 jumps, which I find odd and likely something that can be fixed by some optimization, but seems to hold across a bunch of different benchmarks. But then Direct blows it away. FRM is also unstable for a lot more. So I think it makes sense to just default there.

PR #1231, opened 2025-05-28
Add Global Optimization from BlackBoxOptimizationBenchmarking.jl

ChrisRackauckas:
@jonathanBieler thought you'd want to be aware of this. We're adding it here as an easy way to keep extending it and iterating on it over time. In particular one of the things we want to add is the ability to plot not just iterations but time, because we have some (GPU) parallel solvers in development so how it counts iterations does not work well in comparison to purely serial CPU black box methods, but wall clock time would be a much better comparison.

### DIFF
--- a/benchmarks/GlobalOptimization/blackbox_global_optimizers.jmd
+++ b/benchmarks/GlobalOptimization/blackbox_global_optimizers.jmd
@@ -1,0 +1,162 @@
+---
+title: Black-Box Global Optimizer Benchmarks
+author: Jonathan Bieler, Chris Rackauckas
+---
+
+In this benchmark we will run the [BlackboxGlobalOptimization.jl](https://github.com/jonathanBieler/BlackBoxOptimizationBenchmarking.jl)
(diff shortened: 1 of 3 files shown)


benchmarks/GlobalOptimization/blackbox_global_optimizers.jmd
```diff
@@ -0,0 +1,162 @@
...
+
+## Test one test function (Rastrigin)
+
+```julia
+Δf = 1e-6
+f = test_functions[3]
+
+setup = BenchmarkSetup(NLopt.GN_CRS2_LM(), isboxed=true)
+
+sol = [BBOB.solve_problem(setup, f, 3, 5_000) for in in 1:10]
```
ChrisRackauckas, line 87:
```suggestion
single_setup = BenchmarkSetup(NLopt.GN_CRS2_LM(), isboxed=true)

sol = [BBOB.solve_problem(single_setup, f, 3, 5_000) for in in 1:10]
```

ChrisRackauckas:
We can add PyCMA and SciPy back into the set once https://github.com/SciML/Optimization.jl/issues/918 and https://github.com/SciML/Optimization.jl/issues/917 are done.

PR #1256, opened 2025-05-30
Bump HybirdJumps

ChrisRackauckas:


ChrisRackauckas:
```julia
fig = plot(xlabel = "N", ylabel = "Time");
for (i, algo) in enumerate(algorithms)
    res = results[i]
    plot!(res.t, res.XD[1, :], label = algo.label)
end
title!("2line-Go, AMPA")
```

```
Error: BoundsError: attempt to access 10004-element Vector{Float64} at inde
x [1:21928]
```

@rveltz I think that's a bug in the saving of PiecewiseDeterministicMarkovProcesses.jl? The length of the solution not matching res is odd. Since that's the only issue I'm merging

Issue #163, opened 2020-09-27
State-dependent delay differential equation benchmark


ChrisRackauckas:
@slwu89 donated the following benchmark problem

https://github.com/davewi13/Temperate-Mosquito-DDE
https://gist.github.com/ChrisRackauckas/26b97f963c5f8ca46da19959a9bbbca4

The CR version of the Julia code is a version I updated a little bit. This will be a good benchmark problem for eventual RADAR5-like implementations and such since the Fortran code is right there.

ChrisRackauckas:
Notifying @devmotion @YingboMa

slwu89:
Wow, what a throwback to get a notification here. @sdwfrost, might be interesting to you.

PR #1285, opened 2025-07-07
Add Mosquito state-dependent delay equation benchmark

ChrisRackauckas:
Fixes https://github.com/SciML/SciMLBenchmarks.jl/issues/163


ChrisRackauckas:
Ran locally.

Issue #397, opened 2022-01-18
Bigger non-stiff PDE benchmark


ChrisRackauckas:
https://discourse.julialang.org/t/am-i-going-about-solving-this-pde-system-the-right-way/74621/9

PR #1289, opened 2025-07-07
Add Nonstiff PDE benchmark

ChrisRackauckas:
Fixes https://github.com/SciML/SciMLBenchmarks.jl/issues/397


ChrisRackauckas:
This PDE doesn't actually look non-stiff 😅. It just needs to use sparsity appropriately in the differentiation.

PR #1552, opened 2026-05-12
Fix remaining BayesianInference benchmark issues

ArpanC6:
This PR cleans up and fixes a few lingering issues in the benchmarks/BayesianInference/ folder that were still present after the recent upstream updates.

### What Changed
Fixed inconsistencies and small bugs in DiffEqBayesLorenz.jmd, DiffEqBayesFitzHughNagumo.jmd and Project.toml.

### Why

After syncing with upstream and reviewing the latest commits, I noticed a few problems:

DiffEqBayesLorenz.jmd was importing DiffEqCallbacks twice.
The Stan sampling call in Lorenz was missing : num_samples => 10_000 which made it inconsistent with LotkaVolterra and FitzHughNagumo.
turing_inference and dynamichmc_inference in Lorenz were using @time instead of @btime unlike the other benchmarks.
dynamichmc_inference in Lorenz passed solve_kwargs as a named tuple instead of a Dict which did not match the turing_inference call.
DiffEqBayesFitzHughNagumo.jmd was importing DynamicHMC even though dynamichmc_inference is not used there.
All three benchmarks used SciMLBase.FullSpecialize but SciMLBase was not listed in Project.toml.

### How
Removed the duplicate DiffEqCallbacks import in Lorenz.
Added : num_samples => 10_000 to the Stan call in Lorenz.
Switched @time to @btime in Lorenz benchmarks.
Changed solve_kwargs in dynamichmc_inference to a Dict.
Removed the unused DynamicHMC import from FitzHughNagumo.
Added SciMLBase to [deps] and [compat] in Project.toml.

PR #1553, opened 2026-05-12
Add Bayesian inference benchmark for stochastic Lotka-Volterra SDE

ArpanC6:
### Background

This benchmark was created following a research direction suggested by
Prof. Chris Rackauckas - investigating whether Bayesian pseudo likelihood
methods are strictly better than optimization based MLE for SDE parameter
estimation.

For ODEs the transition density given parameters is exact. For SDEs, the
true transition density $p(X_{t+\Delta t} \mid X_t, \theta)$ is generally
intractable. The Euler Maruyama discretization provides a tractable
pseudo likelihood approximation:

$$p_{\mathrm{EM}}(X_{t+\Delta t} \mid X_t, \theta) \approx \mathcal{N}\!\left(X_t + f(X_t,\theta)\,\Delta t,\ g(X_t,\theta)^2\,\Delta t\right)$$

The key question: given that both Bayesian MCMC and MLE use the same EM
approximation does the Bayesian posterior give strictly more information
than the MLE point estimate?

### What was already in the repo

`benchmarks/BayesianInference/SDELotkaVolterra.jmd` (authored by Chris
Rackauckas) already existed - but that file is a **work-precision diagram**
benchmark comparing SDE solvers (SRIW1, EM, RKMil, SRIW2, SOSRI, SOSRI2)
on strong and weak error. It is **not** a Bayesian inference file.

This PR adds a completely separate benchmark focused on **parameter
estimation and inference** not solver comparison.

### What this PR adds

| File | Change |
|------|--------|
| `benchmarks/BayesianInference/DiffEqBayesLotkaVolterraSDE.jmd` | New file - Bayesian vs MLE benchmark |
| `benchmarks/BayesianInference/Project.toml` | Added `Printf`, `Statistics` to `[deps]` |

### What the benchmark does

Sets up the stochastic Lotka Volterra system with multiplicative noise and
runs two methods head to head on identical synthetic data:

**Bayesian MCMC -  Turing.jl + NUTS**
- Priors placed over all 5 parameters: `a, b, c, d, σ`
- SDE solved inside the Turing model using `remake(prob; p=...)` - SciML
  convention avoids allocating a new `SDEProblem` at every MCMC step
- No fixed random seed inside the model - fixing the seed would collapse
  the stochastic trajectories to a single deterministic path and destroy
  the pseudo-likelihood character
- Full posterior distributions recovered; ESS/s reported

**MLE baseline - Optimization.jl + Nelder-Mead**
- Same pseudo-likelihood minimized to a point estimate
- Gradient free optimizer (Nelder - Mead) used because the SDE solver is
  not reliably differentiable through AD
- Returns only point estimates, no uncertainty quantification

The notebook ends with a side by side comparison table and posterior
histogram plots overlaid with MLE point estimates and true parameter values.

### Key design decisions

| Decision | Reason |
|----------|--------|
| Plain `f(du,u,p,t)` instead of `@ode_def` | Makes `a,b,c,d,σ = p` explicit and unambiguous |
| `remake(prob; p=...)` inside Turing model | SciML convention - avoids allocating new `SDEProblem` at every MCMC step |
| No `seed` inside Turing model | Fixing seed collapses SDE to deterministic path - destroys stochastic pseudo-likelihood |
| Nelder - Mead instead of LBFGS | SDE solver not reliably differentiable through AD; gradient-free is correct here |
| `AutoForwardDiff()` removed from MLE | Same reason as above |
| Separate `@btime` and `@elapsed` cells | Avoids running 1000 MCMC samples twice |
| `chain[param]` not `chain[param].value` | Updated Turing syntax |

### What the benchmark shows

Both methods use the same Euler Maruyama pseudo likelihood so both carry
the same discretization bias. The key difference is how each method handles
that bias.

**MLE** returns a single point estimate locked at the biased value there
is no way to quantify uncertainty around this estimate:

$$\hat{\theta} = \arg\max_\theta \sum_t \log p_{\mathrm{EM}}(X_{t+1} \mid X_t, \theta)$$

**Bayesian MCMC** computes the full posterior:

$$p(\theta \mid X) \propto p(\theta) \prod_t p_{\mathrm{EM}}(X_{t+1} \mid X_t, \theta)$$

The EM bias remains in the likelihood - both methods share it - but the
Bayesian posterior naturally propagates uncertainty around that biased
region. As the noise σ increases or the amount of data decreases, the
posterior widens appropriately. A point estimate fundamentally cannot
capture this.

> **Key insight:** The Bayesian posterior does not remove the EM bias.
> Rather it provides strictly richer information it tells you not only
> where the parameters are likely to be but also how much confidence you
> should place in those estimates.

### Relation to PR #1552

PR #1552 (already open) fixes existing benchmarks in this folder. This PR
is independent - it adds a new benchmark file and does not touch the files
changed in #1552.

### Note on commits

This PR contains two commits. The first (`a244b99`) is from the
`fix/bayesian inference-cleanup` branch which was used as the base
it is the same change as PR #1552. The second (`978c5f6`) is the
actual new benchmark added in this PR.

If preferred I can rebase this branch onto `master` directly so
only the new benchmark commit appears here.


ArpanC6:
WhatsApp Image 2026-05-13 at 12 46 06 AM
WhatsApp Image 2026-05-13 at 12 46 06 AM1

While working on this benchmark I wanted to get a more concrete understanding of the Euler Maruyama (EM) discretization bias. Since the Lotka Volterra SDE has no known exact transition density I used the Ornstein Uhlenbeck process instead where the exact solution is known to compute the bias analytically.
For dX = −θX dt + σ dW the exact transition density has mean Xt · exp(−θΔt) and variance σ²/2θ · (1 − exp(−2θΔt)), while EM uses mean Xt · (1 − θΔt) and variance σ²Δt. Taylor expanding exp(−θΔt) gives mean bias = Xt · θ²Δt²/2 and variance bias = −σ²θΔt², both O(Δt²).

This shows that MLE locks in this O(Δt²) bias at a single point with no way to represent uncertainty around it. In contrast, the Bayesian posterior uses the same biased likelihood but naturally propagates uncertainty around that region. This is exactly why Bayesian inference provides richer information than MLE when fitting SDEs.
I have attached my handwritten calculations above.

ArpanC6:
Heyy, following up on that EM bias thing I mentioned earlier I dug a bit deeper and found something key.

I showed before that the EM variance bias is -σ²θ(Δt)² which is negative, so Var_EM > Var_exact. That means MLE ends up underestimating σ it sees data with less variance than EM expects so it shrinks σ to match.

### Big question: does going Bayesian fix this? Nope not with a weak prior. The posterior still relies on that same biased EM likelihood so the posterior mean underestimates σ too. Bayesian doesn't magically correct the discretization bias from EM.

So yeah "Bayesian is always better" needs some nuance

Uncertainty? Totally posterior gives the full distribution MLE is just a point estimate.

Fixing bias? No way both suffer from the EM likelihood bias.

This is spot on what you said about needing a proof: being Bayesian doesnot make it unbiased. The bias is baked into the likelihood not the inference method.

Curious whether a higher order discretization (like Milstein) 
would reduce this bias and whether that would change the 
Bayesian vs MLE comparison.

ArpanC6:
IMG_0225
IMG_0226
IMG_0227
IMG_0228

You mentioned that a proof would be needed. I have gone through the bias analysis: both EM and Milstein methods introduce an O(Δt2) discretization bias into the likelihood so doing Bayesian MCMC doesn't magically remove that bias. What it does do is give you a richer picture of uncertainty around that biased estimate unlike MLE which gives just a point estimate but it’s still not unbiased. That covers the first part of the argument. The next question is whether the posterior’s uncertainty quantification is actually formally better under some metric.

ChrisRackauckas:
The bias is ignoring truncation error, i.e. assuming dt-> 0.

ArpanC6:
If we have dt kept at a constant value the truncation error will remain; hence the pseudo likelihood will exhibit a bias in terms of dt. The higher order methods such as Milstein scheme would lower the truncation error with constant dt; consequently the pseudo likelihood will improve but not without bias unless dt is zero.

ChrisRackauckas:
But it's not a fixed dt, it's adaptive in general. Assume zero truncation error, the question is if the single shooting scheme within a Bayesian MCMC is unbiased given the conflicting sampling schemes.

ArpanC6:
It never occurred to me from this point of view. Since the random walk is done in an adaptive way, the SDE solver actually introduces randomness at each proposal in the MCMC algorithm; thus, there’s no consistency in the combined method. This needs further thought; does it mean that the reason for inconsistency is the change in the Wiener process?

ChrisRackauckas:
Solving the SDE with the SDE solver is a sample of the solution with one Wiener process. Bayesian MCMC calls a likelihood function and then samples it at different inputs in order to get a posterior. The problem is, if that likelihood function itself is stochastic, what happens? We could, instead of solving the SDE once, we could say solve it 10,000 times and then check that the mean SDE solution matches the data, and that would be a likelihood function. But that would be extremely expensive. But it seems like calling the SDE solver works. Does it actually?

In other words, let Phi(p,W) be the function that takes in parameters p, solves the SDE with Brownian motion W, and then calculates the loss against the data. You know that if you average it out over all W that this likelihood at a given value of `p` is unbiased, and thus MCMC would give the posterior. But if you simply call Phi(p,W) once per MCMC step, if the MCMC process is unbiased and the SDE solver is unbiased, does this compose to giving an unbiased posterior out of the MCMC process? Umberto Picchini has some papers that might be a good starting point to look for this. But also, just empirically, it would be good to start by checking whether this works out in difficult cases.

ArpanC6:
Screenshot (4757)

Andrieu and Roberts (2009) have been studied in connection with the above discussion. The GIMH result indicates that invoking Phi(p,W) at each MCMC iteration is correct under the assumption of unbiasedness of the likelihood estimator. As part of empirical verification of the above I applied a simple OU process where the exact posterior can be determined analytically. The graph enclosed illustrates a comparison between the exact posterior and the posterior using EM single call: for σ there is very good agreement however for θ there is a slight bias with increased variance.

ChrisRackauckas:
We should do this on a larger system of SDEs, like something generated from a Catalyst chemical reaction system. See for example https://docs.sciml.ai/Catalyst/stable/network_analysis/odes/

ArpanC6:
It would also be justified to review Picchini's contributions. In particular as mentioned in both sources the pseudo marginal technique of Andrieu and Roberts (2009) states that one can estimate Φ(p,W) only once during each MCMC update cycle and obtain an unbiased posterior if the likelihood estimator is unbiased. This approach was employed by Picchini (2023) in ABC SMC inference for stochastic differential equations. As far as the Catalyst project is concerned the work from 2025 employs the splitting approach instead of the expectation maximization one and evaluates stochastic Lotka Volterra and Repressilator systems thus resolving the previously raised issue of dt truncation errors. Could we implement the Catalyst benchmark following their approach?

ArpanC6:
ODE with NUTS: The reaction network in Catalyst for the LV model uses the exact parameter values used by Picchini (2025) (θ = 0.5, 0.0025, 0.3; X0 = (100, 100); tspan = [0, 50]). NUTS runs successfully with approximately 80% acceptances; however the chain gets stuck in wrong parameter regions. Many different parameter values lead to almost identical likelihoods meaning that the parameter is unidentifiable. This finding confirms the tendency in the literature of preferring ABC-SMC over NUTS for the LV model.

SDE with pseudo marginal MH: Calculating Φ(p, W) once per each iteration of the chain results in moving the chain and accepting proposals, which is expected from the result for GIMH in Andrieu & Roberts (2009). The main problem is the low acceptance rate caused by the default MH proposals; the situation improves significantly with adjusted proposals.

Incompatibility with ForwardDiff: The Catalyst SDE problem is incompatible with NUTS since Dual numbers produced by ForwardDiff cannot move through the automatically generated functions of SDE. Therefore no gradient based sampling algorithm can be used directly with Catalyst SDE problems.

The next step: I am planning to implement the Lie Trotter splitting scheme introduced in Picchini (2025, Section C.2) as a replacement for EM since EM fails to preserve the oscillations at dt. Do I need to implement the integrator in question myself or does some other method exist?

ArpanC6:
Following up on @ChrisRackauckas's suggestion to test on a larger SDE system and the earlier discussion about EM instability at larger timesteps...

I implemented the Lie Trotter splitting scheme from Jovanovski et al. (2025) Section C.2 for the stochastic Lotka Volterra system.

## Background

The stochastic LV model has a conditionally CIR like structure: when X₂ is fixed X₁ behaves like a CIR-type process, and vice versa. The splitting scheme exploits this structure:

1. Apply the Lamperti transform Z = √X to remove the square-root diffusion
2. Solve the resulting Bernoulli ODE exactly (Φ₁ₐ step)
3. Add the stochastic increment (Φ₁ᵦ step)
4. Compose via Lie Trotter: X₁,next = (Φ₁ᵦ ∘ Φ₁ₐ(√x₁))²

EM has no such structure and can produce negative populations or diverge at larger timesteps.

## Stability comparison
(100 trajectories, θ = (0.5, 0.0025, 0.3), X₀ = (100, 100), T = 50)

| h | Method | Negative populations | Diverged |
|---|--------|---------------------|---------|
| 0.01 | Splitting | 0/100 | 0/100 |
| 0.01 | EM | 2/100 | 0/100 |
| 0.05 | Splitting | 0/100 | 0/100 |
| 0.05 | EM | 5/100 | 0/100 |
| 0.1 | Splitting | 0/100 | 0/100 |
| 0.1 | EM | 5/100 | 0/100 |
| 0.5 | Splitting | 0/100 | 0/100 |
| **0.5** | **EM** | **90/100** | **6/100** |

At h=0.5, EM breaks down completely while the splitting scheme remains stable across all runs, consistent with Figure 4 of Jovanovski et al. (2025).

## Why this matters for inference

EM instability directly corrupts the pseudo-likelihood inside MCMC. Negative populations make the likelihood undefined, forcing either truncation (bias) or proposal rejection (poor mixing). The splitting scheme avoids this entirely especially important when MCMC explores regions far from the posterior.

## Next steps

Would it make more sense to add this as a new benchmark under `benchmarks/BayesianInference/` or first extend the comparison to the Repressilator system?

ChrisRackauckas:
Use the standard methods: SRA3, SOSRI, etc.

ArpanC6:
Tested SRA3 SOSRI and SRIW1 with the Catalyst LV SDE. Only SRA3 works; SOSRI and SRIW1 return an error related to incompatibility with noise type probably because Catalyst produces a non diagonal noise matrix.

SRA3 succeeds: retcode Success; stable positive populations over the entire tspan.

Is it okay to use SRA3 for my benchmark or is there a particular method that can deal with non diagonal noise produced by Catalyst SDEs?

ChrisRackauckas:
it's an additive noise problem?

ArpanC6:
No it is multiplicative noise diffusion terms are sqrt(α·X) sqrt(β·X·Y) sqrt(γ·Y) 
so noise scales with the state. Is there a recommended solver for this case?

ChrisRackauckas:
Oooof not good stuff haha. RKMil

ArpanC6:
RKMil also gives the same noise type incompatibility error as SOSRI and SRIW1. Only SRA3 works with the Catalyst generated noise matrix. Is there a way to convert this to diagonal noise or should we proceed with SRA3?

ChrisRackauckas:
I meant RKMilGeneral. SRA3 only works a high order for additive noise.

ArpanC6:
RKMilGeneral works. Running it now for the benchmark.

ArpanC6:
RKMilGeneral runs fine on its own but inside MCMC it throws an OutOfMemoryError because it allocates Lévy area arrays at every proposal step. Is there a lower memory alternative for non diagonal multiplicative noise in Turing or would it be better to use the EM method with isoutofdomain for this benchmark?

ChrisRackauckas:
not really, looks like the levy area piece needs to get solved. MWE?

ArpanC6:
I have a minimal working example (MWE) using Julia to simulate a stochastic reaction network with Catalyst and StochasticDiffEq and then do Bayesian inference with Turing.
For a short simulation tspan=10 with 5 samples everything works fine. But when I try a longer simulation tspan=50 with 200 samples I get an OutOfMemoryError.
The problem seems to be with RKMilGeneral(). This solver allocates Lévy area arrays at every proposal and for long time horizons and many samples those arrays accumulate and blow up memory usage.

Here is a simplified version of my code:
```julia
using Catalyst, StochasticDiffEq, Turing, Distributions, LinearAlgebra

rn = @reaction_network begin
    α, X --> 2X
    β, X + Y --> 2Y
    γ, Y --> ∅
end

u0 = [:X => 100.0, :Y => 100.0]
p_true = [:α => 0.5, :β => 0.0025, :γ => 0.3]

prob_short = SDEProblem(rn, u0, (0.0, 10.0), p_true)
sol_data = solve(prob_short, RKMilGeneral(), dt=0.1, saveat=1.0)
obs = Array(sol_data)

@model function lv_mwe(obs, prob)
    α ~ truncated(Normal(0.5, 0.5), 0, 2)
    p = [:α => α, :β => 0.0025, :γ => 0.3]
    _prob = remake(prob, p=p)
    sol = solve(_prob, RKMilGeneral(), dt=0.1, saveat=1.0)
    sol.retcode != ReturnCode.Success && (Turing.@addlogprob! -Inf; return)
    predicted = Array(sol)
    for i in 2:size(obs,2)
        obs[:,i] ~ MvNormal(predicted[:,i], 10.0*I)
    end
end

sample(lv_mwe(obs, prob_short), MH(), 5)  #OK

prob_long = SDEProblem(rn, u0, (0.0, 50.0), p_true)
sol_data2 = solve(prob_long, RKMilGeneral(), dt=0.1, saveat=1.0)
obs2 = Array(sol_data2)
sample(lv_mwe(obs2, prob_long), MH(), 200)  #OutOfMemoryError
```

In short RKMilGeneral() is fine for short simulations but for longer time spans and more samples the memory usage explodes because it repeatedly allocates Lévy area arrays for each proposal.

ChrisRackauckas:
no, isolate it to no catalyst no turing.

ArpanC6:
Isolated to no Catalyst no Turing. Pure StochasticDiffEq with non diagonal noise 200 solves with GC.gc() after each completes without OOM. So the allocation in LevyArea itself is not the root cause; GC can reclaim it normally. The issue is that Turing's model evaluation holds references that prevent GC from freeing the LevyArea arrays between proposals.

ArpanC6:
Correction to my previous comment further tested plain SDEProblem (no Catalyst) with RKMilGeneral inside Turing MH() this also gives OutOfMemoryError at levyarea/milstein.jl:25. So the issue is RKMilGeneral + Turing together not Catalyst. Plain loop of 200 solves without Turing completes fine even without GC.gc().

ArpanC6:
I tested EM() using the isoutofdomain=(u,p,t) -> any(x -> x < 0, u) check within Turing’s MH() as a workaround. The process completed 200 iterations successfully without any out-of-memory errors. Some instability warnings appear when MCMC proposals enter unstable parameter regions but these are properly handled through the domain check. This approach could serve as a practical alternative for the benchmark while the memory issues with RKMilGeneral + Turing are investigated.

ArpanC6:
I ran the updated benchmark locally and here is what I found:

1) NUTS with SRIW1: The chain got completely stuck only 1 unique value in 1000 samples. The stochasticity in the SDE prevents proper gradient computation.

2) MH with default proposals: Slightly better but still very low acceptance only 11 unique values in 1000 samples.

3) MH with tuned proposals (σ = 0.1): Still stuck mostly because the SRIW1 solver fails repeatedly at these parameter values.

4) MLE (Nelder Mead): Works well converges in ~2.6s and gives reasonable estimates.

So his confirms our earlier discussion: the LV SDE is really tough for standard MCMC methods. Using a pseudo marginal MH approach (i.e., just the plain SDEProblem without a fixed seed) might actually be more suitable for the benchmark.

Do you want me to go ahead and switch to that approach for the benchmark?

ChrisRackauckas:
> NUTS with SRIW1: The chain got completely stuck only 1 unique value in 1000 samples. The stochasticity in the SDE prevents proper gradient computation.

No, check what you're doing. You're getting LLM hallucinations.

ArpanC6:
You are right I apologize for the earlier explanation. Looking back at the diagnostics the acceptance rate was 1.0 but the chain had only a single unique value. This suggests that NUTS is failing to compute valid gradients through the SDE solver likely because SRIW1 is not compatible with automatic differentiation. I should have checked this properly before posting.

ArpanC6:
I ran the diagnostics properly this time. We got 50 unique values no numerical errors and an acceptance rate of 0.57 so NUTS is computing gradients just fine. The real issue is that the mean tree depth is 10.0 hitting the maximum at every step. This suggests the posterior is too flat or multimodal for NUTS to explore efficiently which fits with the identifiability issues in the LV model. My earlier explanation about AD incompatibility was incorrect.

ArpanC6:
??

ChrisRackauckas:
It's almost certainly multimodal

ArpanC6:
Opened a cleaner PR at #1569 addressing the review feedback. Previous unrelated commits have also carried over will rebase to a single commit if needed.

ArpanC6:
Screenshot (4871)
Screenshot (4873)

I ran 200 NUTS samples locally. The chain has 200 unique values so it’s mixing well. The average tree depth is 9.97 hitting the maximum at every step. The histogram shows a peaking around 1.2-1.4 and c around 2.0-2.2 while the true values are 1.5 and 3.0, close by. The high tree depth suggests the posterior geometry is tricky for NUTS which could mean it’s multimodal or has a very curved shape.

ArpanC6:
??

ArpanC6:
Could you please let me know if I made any mistakes in what I did..

ArpanC6:
Hlw @ChrisRackauckas could you let me know what you'd like me to do next on this PR?

ArpanC6:
I have opened a cleaner version of this benchmark at #1569 addressing all the previous feedback. Could you take a look when you get a chance??

ArpanC6:
@ChrisRackauckas I investigated the sampler issue further. Results

NUTS without seed: 200 divergent transitions, unique samples = 1
NUTS with fixed seed: tree depth = 1023, unique samples = 200 (but pseudo-likelihood is wrong)
MH with tuned proposals: SDE solver aborts at unstable parameter regions, unique samples = 1

The core issue is that the stochastic likelihood causes both NUTS and MH to fail. Would pseudo marginal MH with a more stable solver like EM + isoutofdomain be the right direction here???

ChrisRackauckas:
might be easiest to jump on a call at this point.

ArpanC6:
Sure I am available for a call. Please let me know what time works for you??

ArpanC6:
I need to correct my earlier diagnosis about the solver issue.

I spent today running proper isolated tests on both the original 
benchmark setup and a larger one and I realized I was wrong about 
it being a memory leak.

With RKMilGeneral + Turing on the larger setup the run finished in 
about 57 seconds for 200 samples, but there were only 6 unique 
samples. I also saw around 150 "dt <= dtmin aborting. Solution is 
unstable" warnings.

It looks like the solver is running into unstable parameter regions 
during sampling and aborting. When that happens, the solver returns 
a failure Turing assigns a log probability of -Inf the proposal 
gets rejected and the chain eventually gets stuck.

I then tried the same setup using EM with isoutofdomain, and it 
worked much better. There were no solver warnings and the chain 
sampled normally.

So I don't think this is actually a memory issue. It seems to be a 
solver stability issue when MCMC explores difficult parameter 
regions.

If you agree i will update PR #1569 to use EM with isoutofdomain.

ArpanC6:
Closing this in favor of #1569 which contains the rebased and updated fix.

PR #1554, opened 2026-05-13
Fix DiffEqBayesLotkaVolterra: merge split imports, fix sample_kwargs type, switch to @elapsed

ArpanC6:
### What Changed
Fixed three issues in benchmarks/BayesianInference/DiffEqBayesLotkaVolterra.jmd that were exactly the same as the ones we fixed in #1552 for Lorenz and FitzHugh-Nagumo.

### Why
While working on #1552, I noticed that DiffEqBayesLotkaVolterra.jmd had the same problems that we already cleaned up in the other two files. So I went ahead and fixed them here too

Split using blocks The imports were unnecessarily split into two separate Julia code cells. Merged them into one for better readability.
@time vs @btime The Stan inference was using @time while Turing and DynamicHMC used @btime. Made it consistent by switching to @elapsed (since Stan’s warmup is part of the total runtime and we don’t want to exclude it). Added a small comment explaining the choice.
sample_kwargs as Dict Was passing a Dict instead of a NamedTuple. Changed it to (delta = 0.85, num_samples = 10_000) to follow Julia conventions and stay consistent with the rest of the codebase.

### How

Combined the two using blocks into one
Replaced @time with @elapsed (and stored the result)
Changed the Dict to a NamedTuple for sample_kwargs

This PR basically brings DiffEqBayesLotkaVolterra.jmd in line with the improvements already made in #1552.

### DIFF
--- a/benchmarks/BayesianInference/DiffEqBayesLotkaVolterra.jmd
+++ b/benchmarks/BayesianInference/DiffEqBayesLotkaVolterra.jmd
@@ -8,9 +8,6 @@ author: Vaibhav Dixit, Chris Rackauckas
 
 ```julia
 using DiffEqBayes, StanSample, DynamicHMC, Turing
-```
-
-```julia
 using Distributions, BenchmarkTools, StaticArrays
 using OrdinaryDiffEq, RecursiveArrayTools, ParameterizedFunctions
 using Plots, LinearAlgebra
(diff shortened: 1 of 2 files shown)


ArpanC6:
While fixing these issues I also realized that using `@elapsed` for Stan actually makes more sense than `@btime`. The reason is that Stan’s warmup isn’t a one time cost it happens every time you run it. If you use `@btime`, it runs Stan multiple times and averages out the warmup which ends up underestimating the true cost. Using `@elapsed` with a single run gives a more accurate picture of Stan’s actual wall time.
This also lines up with how the direct Turing block works which uses `@elapsed` as well (so it can pass the elapsed time to `display_ess_per_sec`)

PR #1569, opened 2026-05-18
Add Bayesian inference benchmark for stochastic Lotka-Volterra SDE

ArpanC6:
This benchmark compares Bayesian MCMC (Turing + MH) with MLE (LBFGS + AutoForwardDiff) for estimating parameters of the stochastic Lotka Volterra system using the Euler-Maruyama pseudo-likelihood.

What's changed from the previous version

Replaced @ode_def with a plain f(du, u, p, t) function so the parameters a, b, c, d, σ are clear.
Removed the fixed seed in the Turing model - using a fixed seed made the SDE behave deterministically which breaks the pseudo-likelihood.
Switched MLE optimizer to LBFGS with AutoForwardDiff. An MWE confirms ForwardDiff successfully differentiates through SRIW1() for this problem, and LBFGS converges in ~15 iterations vs NelderMead failing after 1000.
Removed unused imports StaticArrays and ParameterizedFunctions.
Fixed chain indexing using vec(Array(chain[param])).
Cleaned up empty code blocks.

Note: This PR only has the new benchmark commit. The previous PR (#1553) also had an unrelated commit from another branch.

### DIFF
--- a/benchmarks/BayesianInference/DiffEqBayesLotkaVolterra.jmd
+++ b/benchmarks/BayesianInference/DiffEqBayesLotkaVolterra.jmd
@@ -8,9 +8,6 @@ author: Vaibhav Dixit, Chris Rackauckas
 
 ```julia
 using DiffEqBayes, StanSample, DynamicHMC, Turing
-```
-
-```julia
 using Distributions, BenchmarkTools, StaticArrays
 using OrdinaryDiffEq, RecursiveArrayTools, ParameterizedFunctions
 using Plots, LinearAlgebra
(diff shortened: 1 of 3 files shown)


ArpanC6:
@ChrisRackauckas Following up on my earlier diagnosis I have gone ahead and pushed the fix

Switched to EM(dt=0.01) with isoutofdomain instead of SRIW1() which stops the dt <= dtmin aborts when MCMC explores unstable parameter regions (like negative populations).

Using MH() instead of NUTS() since NUTS was hitting the max tree depth at 10.0 on this multimodal posterior.

Using NelderMead() for MLE instead of LBFGS() gradient-free makes more sense here since the SDE solver isn't reliably differentiable through AD.

Everything is rebased on the latest master. CI should be clean now feel free to take a look whenever you get a chance.

ChrisRackauckas:
> Using NelderMead() for MLE instead of LBFGS() gradient-free makes more sense here since the SDE solver isn't reliably differentiable through AD.

That doesn't make sense. MWE?

ArpanC6:
> > Using NelderMead() for MLE instead of LBFGS() gradient-free makes more sense here since the SDE solver isn't reliably differentiable through AD.
> 
> That doesn't make sense. MWE?

You're right I ran an MWE and ForwardDiff.gradient does work with SRIW1() on this simple loss. I was wrong about AD being fundamentally broken here.

PR #1606, opened 2026-07-02
Fix preconditioner precs interface for v7 (#1601)

### LINKED ISSUE #1601: Preconditioner setup only partially upgraded
**Describe the bug 🐞**

Preconditioner setup changed in v7 from `myprec(W, du, u, p, t, newW, Plprev, Prprev, solverdata)` to `myprec(newW, p)`. Furthermore, a `SciMLOperators.WOperator` is passed despite `concrete_jac = true` being passed into the KenCarp solvers.

**Expected behavior**

Concrete Jac and the benchmarks should work without manual intervention.

**Minimal Reproducible Example 👇**

```julia
using SciMLBenchmarks
SciMLBenchmarks.weave_file(joinpath(pkgdir(SciMLBenchmarks),"benchmarks","SimpleHandwrittenPDE"),"burgers_fdm_wpd.jmd")
```

**Error & Stacktrace ⚠️**

```
ERROR: MethodError: no method matching (::Main.var"##WeaveSandBox#279".IncompleteLUPreconBuilder)(::SciMLOperators.WOperator{…}, ::Tuple{…})
The object of type `Main.var"##WeaveSandBox#279".IncompleteLUPreconBuilder` exists, but no method is defined for this combination of argument types when trying to treat it as a callable object.
```

**Environment (please complete the following information):**

Master as-is.

jitendravjh:
Fixes #1601. The `precs` interface changed in v7 from the 9-argument `(W, du, u, p, t, newW, Plprev, Prprev, solverdata)` to the 2-argument `(newW, p)`, so the hand-written preconditioner builders in the benchmarks were failing with a `MethodError` on the old signature.

This migrates all of them to `(newW, p)` and returns `I` instead of `nothing` for the right preconditioner, since the solver now applies it via `ldiv!` and `nothing` has no method. `convert(AbstractMatrix, newW)` materializes the passed `WOperator` to its sparse matrix, so `concrete_jac = true` keeps working.

Updated 12 builders across `SimpleHandwrittenPDE/burgers_fdm_wpd.jmd`, `SimpleHandwrittenPDE/allen_cahn_fdm_wpd.jmd`, `StiffODE/Bruss.jmd`, `Bio/BCR.jmd`, and `Bio/fceri_gamma2.jmd`.

Checked locally by running the affected builders on the real problems: the four SimpleHandwrittenPDE preconditioners solve the burgers problem with `KenCarp4`, and `KenCarp47` solves with the ILU builder. The Bio problems use a sparse Jacobian, so the `ilu(convert(...))` path is unchanged from before.

### DIFF
--- a/benchmarks/Bio/BCR.jmd
+++ b/benchmarks/Bio/BCR.jmd
@@ -135,13 +135,9 @@ function precilu(z, r, p, t, y, fy, gamma, delta, lr)
     ldiv!(z, preccache[], r)
 end
 
-function incompletelu(W, du, u, p, t, newW, Plprev, Prprev, solverdata)
-    if newW === nothing || newW
-        Pl = ilu(convert(AbstractMatrix, W), τ = τ2)
-    else
-        Pl = Plprev
-    end
-    Pl, nothing
+function incompletelu(newW, p)
+    Pl = ilu(convert(AbstractMatrix, newW), τ = τ2)
+    Pl, I
 end;
 ```
 
--- a/benchmarks/Bio/fceri_gamma2.jmd
+++ b/benchmarks/Bio/fceri_gamma2.jmd
@@ -126,13 +126,9 @@ function precilu(z, r, p, t, y, fy, gamma, delta, lr)
 end
 
 const τ2 = 5
-function incompletelu(W, du, u, p, t, newW, Plprev, Prprev, solverdata)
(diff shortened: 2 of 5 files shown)


ChrisRackauckas:
Sorry I had the exact same migration in flight, and it just merged shortly after you put this up.

PR #1612, opened 2026-07-15
Regenerate Testing manifest with Julia 1.10

ChrisRackauckas-Claude:
Please ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- Regenerate `benchmarks/Testing/Manifest.toml` with Julia 1.10.11, matching the Julia 1.10 package-test job.
- Remove the Julia 1.11-only `StyledStrings` stdlib package entry and regenerate the other Julia-version-specific stdlib records through Pkg rather than editing the manifest by hand.
- Keep the fix manifest-only. The existing `weave_file` test directly covers the failure and passes with the regenerated environment.

## Root cause

The Testing manifest was regenerated with Julia 1.11.9 in `cbafe70c10768754c99856915a72e094d0b6042a`, but `.github/workflows/test.yml` still runs Julia 1.10. Julia 1.10 can instantiate that manifest with a warning, then `Pkg.build()` fails because `StyledStrings` is a Julia 1.11 stdlib and has no package path on Julia 1.10.

A `git bisect run` using `Pkg.instantiate(); Pkg.build()` found:

- first bad: `cbafe70c10768754c99856915a72e094d0b6042a`
- good parent: `6abeb2ca2c7b0dc2a2dce5f1c811950f5fc87e5f`

This follows the prior Testing-environment fix in #1497, which likewise regenerated this manifest with Julia 1.10 to match the package-test job.

## Local verification

- Before the change, Julia 1.10.11 `Pkg.test()` reproduced `Failed to find path for package StyledStrings` in `weave_file` and exited 1.
- After the change, Julia 1.10.11 `Pkg.test()` passed: Explicit Imports 2/2 and `weave_file` 1/1.
- Julia 1.11.9 `Pkg.instantiate(); Pkg.build()` for `benchmarks/Testing` exited 0. Pkg emits the expected warning that the manifest was resolved with Julia 1.10.11.
- Runic checked all 10 tracked Julia source files successfully.
- `git diff --check` passed.

### DIFF
--- a/benchmarks/Testing/Manifest.toml
+++ b/benchmarks/Testing/Manifest.toml
@@ -1,8 +1,8 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.11.9"
+julia_version = "1.10.11"
 manifest_format = "2.0"
-project_hash = "c00133cf365b9e2da39b2c55a741e1626ba8a1d9"
+project_hash = "68ba5be0a107f936888c42eedc0265ec346c3e0f"
 
 [[deps.AliasTables]]
 deps = ["PtrArrays", "Random"]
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Rebased this one-file manifest refresh onto current `master` `1c15d9ad` and
updated the existing branch with a guarded lease after fetching both remotes
and confirming that its only remote-only commit was our original
`b5ac778b`. The exact new head is `8b9ff5f51fa9253aeb8566f888adb4cb50bcf48f`.

Post-rebase local verification used Julia 1.10.11 with both the depot and
temporary directory confined to the workspace:

- `Pkg.test()` passed the official package suite: Explicit Imports **2/2** and
  `weave_file` **1/1**. The nested Testing environment instantiated and wove
  all four chunks successfully. Julia emitted the existing IJulia-extension
  method-overwrite precompile warning, then continued through both passing
  testsets.
- `benchmarks/Testing/Manifest.toml` parses as Julia **1.10.11** with 202
  dependency entries; SHA-256
  `73a50e18ef512b951c0a64a76c780f6ff5afe1e82b9a56795b14e12f469ff830`.
- Repository-wide Runic 1.7 `--check .` exited 0.
- `git diff --check origin/master...HEAD` and the final clean-worktree check
  passed.

No test or assertion was skipped, disabled, weakened, or silenced.

PR #1613, opened 2026-07-15
Fix PSO benchmark : skip unstable BBOB functions

AdityaPandeyCN:
Pin CUDA runtime via LocalPreferences extras and exclude f4, f7, f10 which cause segfaults or illegal memory access on GPU.

## Checklist

- [ ] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API
  
## Additional context
The PR also adds updated Manifest.toml

PR #1513, opened 2026-03-21
Add Parallel Particle Swarm Optimizer Benchmarks

AdityaPandeyCN:
## Checklist

- [ ] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API
  
## Additional context

Adds GPU PSO benchmark page comparing ParallelParticleSwarms.jl variants (SerialPSO, ParallelPSOKernel sync/async, ParallelSyncPSOKernel, HybridPSO-LBFGS) across CPU and GPU backends on Rosenbrock test problems.

PR #1614, opened 2026-07-17
Use descriptive SerialPSO benchmark alias

ChrisRackauckas-Claude:
Please ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- Rename the local `SerPSO` benchmark alias to the descriptive `SerialPSOAlgorithm`.
- Update the committed rendered HTML to match the benchmark source.
- Leave `.typos.toml` and the spell-check workflow unchanged, so the underlying spelling problem is fixed rather than allow-listed or ignored.

## Root cause

`crate-ci/typos` 1.48.0 tokenizes the `Ser` prefix in `SerPSO` as a misspelling and reports both the alias declaration and its use. Both occurrences were added in `0498ef2519fdc2f60a1a2d78502f971bdf0b3448` (`Add Parallel Particle Swarm Optimizer Benchmarks (#1513)`).

An exact local reproduction on clean `master` exited 2 with the same two diagnostics as the failing spell-check job. The parent `b490985f38204c12ebc426e3293af855fa536536` exited 0. A `git bisect run` identified `0498ef2519fdc2f60a1a2d78502f971bdf0b3448` as the first bad commit.

## Local verification

- `typos-cli 1.48.0`: `typos .` exits 0 on this branch.
- Runic 1.7.0: all 10 tracked Julia files pass a repository-wide `--check --verbose` run.
- `git diff --check origin/master...HEAD` exits 0.
- Julia 1.10.11 `Pkg.test()` on unmodified `master` and this branch reaches Explicit Imports (2/2) and then reproduces the unrelated `StyledStrings` Testing-manifest failure already bisected and fixed separately in #1612.
- In a temporary validation worktree with only #1612's manifest commit layered on this commit, Julia 1.10.11 `Pkg.test()` exits 0: Explicit Imports 2/2 and `weave_file` 1/1.

This PR is intentionally separate from #1612 and from the downgrade-workflow branch. Open PSO follow-up #1613 touches the same benchmark source but does not rename or remove `SerPSO`, so it is not a duplicate of this fix.


ChrisRackauckas-Claude:
CI triage after the conservative timer:

- `typos` passes, confirming the two `SerPSO` diagnostics are fixed.
- Runic, Runic Suggestions, and changed-benchmark detection pass.
- The PSO benchmark job resolves an unpinned CUDA 13.2 runtime on the repository's Tesla V100 (compute capability 7.0), then fails because CUDA 13 requires compute capability 7.5 or newer.

This branch changes only the local algorithm alias and its rendered output; it does not change the benchmark environment or CUDA selection. The existing #1613 is the focused environment fix: it pins CUDA 12.9 for the V100, and its two latest benchmark workflow runs pass. I have kept this spelling PR separate and have not suppressed the GPU failure.

PR #1616, opened 2026-07-17
Require MbedTLS 1.1.1 for downgrade tests

ChrisRackauckas-Claude:
## Summary

- declare MbedTLS as a direct dependency so downgrade resolution can enforce its usable lower bound
- require MbedTLS 1.1.1, the first tested release that completes HTTP precompilation in this dependency graph

This PR depends on #1612, which refreshes the checked-in manifests used by the benchmark test path.

## Local verification

- Julia 1.10 locked downgrade resolution on top of #1612 selected HTTP 1.11.0 and MbedTLS 1.1.1
- the locked `Pkg.build()` exited 0 and strict `Pkg.test()` passed Explicit Imports 2/2 and the Weave test 1/1
- MbedTLS 1.1.0 spent more than 60 minutes at 99.8% CPU compiling HTTP before being interrupted; the equivalent isolated MbedTLS 1.1.1 control completed in 17 seconds
- the previous MbedTLS 0.7 floor also failed to finish within a two-hour build window
- Julia 1.12 latest-dependency build selected MbedTLS 1.1.10 and exited 0
- Julia 1.12 strict testing passed Explicit Imports 2/2, then reproduced an unrelated clean-master nested-manifest failure (`Failed to find path for package MbedTLS_jll`); that clean-main regression is being bisected separately as required
- Julia Runic 1.7.0 check, `git diff --check`, and Project.toml dependency/compat assertions passed

## Review

Ignore this PR until it has been reviewed by @ChrisRackauckas.


ChrisRackauckas-Claude:
Investigation scratchpad for commit c1b7e894:

1. Reproduced the downgrade build pathology with the historical transitive MbedTLS floor. MbedTLS 0.7 did not finish within two hours.
2. Tested the apparent major-version boundary. MbedTLS 1.1.0 still left the HTTP precompile worker at 99.8% CPU after 60 minutes, so a broad `1.1` floor was not sufficient.
3. Ran an isolated HTTP 1.11.0 + MbedTLS 1.1.1 control. It precompiled in 17 seconds.
4. Added MbedTLS directly with floor 1.1.1, ran the same minimum-version resolver used by downgrade CI on top of #1612, built the locked environment, and ran the actual package tests. Explicit Imports passed 2/2 and Weave passed 1/1.
5. Tested current dependencies on Julia 1.12. The build selected MbedTLS 1.1.10 and exited 0. The test reached an unrelated nested-manifest error that was then reproduced unchanged on unmodified master; a separate required clean-main bisect/fix investigation is active.
6. Re-ran Julia Runic 1.7.0, TOML assertions, and Git whitespace/diff checks before committing. The branch was fetched and confirmed 0 commits behind master before its non-force push.

ChrisRackauckas-Claude:
CI note: the Spell Check job's two failures are the unchanged `SerPSO` findings on master. They are fixed separately by #1614; commit c1b7e894 changes only Project.toml.

ChrisRackauckas-Claude:
CI triage after the conservative timer:

- Runic and Runic Suggestions pass.
- Spell Check reports only the two unchanged `SerPSO` findings from clean `master`; their first-bad commit and fix are handled separately in #1614.
- Julia Tests passes Explicit Imports 2/2, then fails while instantiating the unchanged `benchmarks/Testing/Manifest.toml`: Julia 1.10 cannot locate the Julia 1.11 `StyledStrings` stdlib entry. That clean-base failure was reproduced and bisected separately, and #1612 regenerates the nested manifest with Julia 1.10.

Commit `c1b7e894` changes only the root `Project.toml`, so neither failure is caused by the MbedTLS floor. I have left this PR focused and have not hidden either failure.

PR #1618, opened 2026-07-17
Refresh nested manifests before benchmark builds

ChrisRackauckas-Claude:
> Ignore this PR until it has been reviewed by @ChrisRackauckas.

Depends on #1612 for the Julia 1.10-compatible `Testing` manifest.

## Summary

- Resolve a nested benchmark manifest before instantiation when the running Julia minor is the same as or newer than the manifest's Julia minor.
- Preserve manifests produced by a newer Julia when running on an older Julia, whose resolver may not know the newer stdlibs.
- Add a regression test using a stale minimal manifest.

## Root cause

Julia 1.12 no longer bundles `MbedTLS_jll`, while the checked-in Julia 1.11 `Testing` manifest records it as a stdlib without a source tree. `Pkg.instantiate()` therefore leaves no package path for the subsequent `Pkg.build()` call.

A full bisect found `cc4abfcee7f55d5be8be72060f382c1de5d1742f` as the first bad commit because it introduced and activated the nested `Testing` environment. Resolving the manifest on same-or-newer Julia versions converts historical stdlibs such as `MbedTLS_jll` to normal registered packages before building.

## Verification

- Julia 1.12: full `Pkg.test()` passes on this branch, including ExplicitImports, the stale-manifest regression, and the real woven benchmark test.
- Julia 1.10: full `Pkg.test()` passes when stacked on #1612.
- Runic 1.7.0: full-repository check passes.

### DIFF
--- a/src/SciMLBenchmarks.jl
+++ b/src/SciMLBenchmarks.jl
@@ -16,6 +16,13 @@ macro subprocess(ex, wait = true)
     end
 end
 
+function manifest_julia_version(folder)
+    manifest_file = joinpath(folder, "Manifest.toml")
+    isfile(manifest_file) || return nothing
+    version_match = match(r"^julia_version\s*=\s*\"([^\"]+)\""m, read(manifest_file, String))
+    return isnothing(version_match) ? nothing : VersionNumber(only(version_match.captures))
+end
+
 function weave_file(folder, file, build_list = (:script, :github))
     Weave.set_chunk_defaults!(:error => false)
     target = joinpath(folder, file)
@@ -25,6 +32,13 @@ function weave_file(folder, file, build_list = (:script, :github))
         @info("Instantiating", folder)
         Pkg.activate(folder)
         withenv("JULIA_PKG_PRECOMPILE_AUTO" => "0") do
+            manifest_version = manifest_julia_version(folder)
+            # Older Julia releases cannot resolve stdlibs pinned by a newer manifest.
+            if manifest_version === nothing ||
+                    (VERSION.major, VERSION.minor) >=
+                    (manifest_version.major, manifest_version.minor)
+                Pkg.resolve()
+            end
             Pkg.instantiate()
         end
         Pkg.build()
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -4,6 +4,29 @@ using SciMLBenchmarks, Test
     include("explicit_imports.jl")
 end
 
+@testset "weave_file resolves stale manifests" begin
+    mktempdir() do folder
+        write(
+            joinpath(folder, "Project.toml"),
+            """
+            [deps]
+            Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+            """,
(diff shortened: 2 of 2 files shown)


ChrisRackauckas-Claude:
Investigation scratchpad

Task prompt: independently reproduce the clean-master Julia 1.12 `Pkg.test()` failure, bisect it, fix the underlying nested-environment behavior without skipping tests, and verify the actual test locally.

- Clean base: `0498ef2519fdc2f60a1a2d78502f971bdf0b3448`.
- Independent reproducer: Julia 1.12 with an isolated depot, activating `benchmarks/Testing` and running `Pkg.instantiate(); Pkg.build()`.
- Reproduction result: exit 1, `Failed to find path for package MbedTLS_jll`.
- Bisect range: clean master bad, `cc4abfce^` good; automated predicate ran the public `Pkg.instantiate(); Pkg.build()` path and classified only the exact `MbedTLS_jll` path error as bad.
- Bisect result: first bad `cc4abfcee7f55d5be8be72060f382c1de5d1742f` (`Move test.jmd into its own project folder`). That commit introduced the nested environment and made the test activate its manifest.
- First repair probe: resolving before instantiation on Julia 1.12 added a source tree for the historical `MbedTLS_jll` stdlib and allowed `Pkg.build()` to complete.
- Compatibility iteration: unconditional resolution is invalid when an older Julia reads a newer manifest; Julia 1.10 cannot resolve the manifest's Julia 1.11 `StyledStrings` stdlib pin. The implementation therefore resolves only on the same or a newer Julia minor.
- Added regression: a temporary project with a deliberately stale manifest must be resolved and instantiated by `weave_file`.
- Julia 1.12 full test result: Explicit Imports 2/2, stale-manifest regression 2/2, real `weave_file` 1/1; `Pkg.test()` passed.
- Julia 1.10 validation stacked on #1612: Explicit Imports 2/2, stale-manifest regression 2/2, real `weave_file` 1/1; `Pkg.test()` passed.
- Formatting: Runic 1.7.0 full-repository check passed.

No tests were skipped, disabled, loosened, or silenced.

ChrisRackauckas-Claude:
CI note: the immediate Spell Check failure is the two pre-existing `SerPSO` findings on clean master. It is unrelated to this two-file diff and is fixed separately by #1614.

ChrisRackauckas-Claude:
Post-timer CI review:

- **Runic Format** and **Runic Suggestions** pass.
- **Julia Tests:** Explicit Imports (2/2) and the new stale-manifest regression (2/2) pass. The only failure is the pre-existing Julia 1.10 read of the Julia 1.11 `Testing` manifest: `Could not locate source code for StyledStrings`. This is the exact prerequisite addressed by #1612; the full Julia 1.10 `Pkg.test()` passes locally when this branch is stacked on #1612.
- **Spell Check:** the exact two clean-`master` `SerPSO` findings remain; they are addressed by #1614.

No branch-local CI failure was found, so no additional change was pushed.

PR #1629, opened 2026-08-06
Add multirate split ODE work-precision benchmarks

singhharsh1708:
Adds work-precision benchmarks for the OrdinaryDiffEqMultirate solvers (MREEF, MRAB, MIS, and the explicit and solve-decoupled implicit MRI-GARK families) to the SimpleHandwrittenPDE suite. The methods run inside the existing Burgers, Kuramoto-Sivashinsky and Allen-Cahn notebooks against each file's own references plus a consistent cross-family reference set in every multirate plot: Tsit5, Rodas5P, FBDF, KenCarp4 and ETDRK4. No new files.

On Burgers, adaptive Tsit5 is the fastest method through the middle of the range (stability capped at half a millisecond while its error falls to 1.3e-7), and the multirate members' territory is the tight end: MREEF reaches 2.2e-10 in about 10 ms, errors no adaptive reference touches, with only fixed-step ETDRK4 going deeper at 25x the cost. The prose attributes that tight-end margin to step economics, not multirate structure:

Burgers multirate

Kuramoto-Sivashinsky is decided by the cross-family references: FBDF dominates outright (7 to 9 ms per solve across the sweep, one to two orders cheaper than anything else at matched error), and the explicit multirate members are not competitive there, which the notebook states plainly:

KS explicit multirate

The implicit MRI-GARK members on the reversed split beat the KenCarps making the same term assignment by 30 to 300x at matched error, but the notebook caps the claim: FBDF and Tsit5 on the joined problem beat the best multirate member by one to two orders once the split constraint is dropped. One stated wart: ESDIRK46a's error stops responding at the tightest tolerances:

KS implicit multirate

Allen-Cahn is the null case: no rate separation, Tsit5 fastest at every error it reaches, the micro substepping is pure overhead, and the section documents the boundary of the family's applicability:

Allen-Cahn multirate

All three notebooks were executed end to end (Burgers and Allen-Cahn fully in CI, KS through all multirate blocks in CI with the untouched pre-existing low-tolerance block verified separately); the figures above are from those runs and the benchmark server regenerates them from the jmd. Writing the cross-family blocks also surfaced that a direct call of a SplitFunction corrupts prob.u0 through its aliased internal cache, which will be filed upstream separately. The linearly implicit extrapolated member (MREIL) follows once OrdinaryDiffEqMultirateImplicit is registered.

## AI Disclosure

Claude assisted with this work.


ChrisRackauckas:
why is this a separate folder? Why not in the handwritten pdes?

singhharsh1708:
moved it into SimpleHandwrittenPDE as multirate_wpd.jmd, folder deleted. added the seven missing deps to that env (Multirate pinned 2.7 in compat, the resolver initially picked 2.6 which lacks ESDIRK46a) and re-ran all 12 blocks there, reference agreements at 1e-14.

ChrisRackauckas:
But why these other problems instead of the standard PDEs? I don't see a reason for a new file like this.

singhharsh1708:
moved them into the standard notebooks and deleted the file. reran all three end to end, description updated with the new figures. the implicit mri-gark members genuinely win on ks, the explicit ones tie cnab2 there, and allen cahn stays as the null case where tsit5 wins.

ChrisRackauckas:
The cross family checks aren't consistent here. Tsit5, Rodas5P, FBDF (NordsieckBDF), KenCarp4, ETDRK4, and the new methods would get a good feel for how it's done. I'll probably do some passes to improve this a bit but it would be nice to at least see these plots for now.

> the null case where tsit5 wins.

It has an accuracy limit though, which is the expected issue.

singhharsh1708:
added the cross family set to every multirate plot and reran everything. the honest version: fbdf owns ks outright, tsit5 owns the middle of burgers and allen cahn, and the multirate members only earn their cost below 1e-8 on burgers or under a fixed split constraint on ks. also found a nasty one while debugging: calling a SplitFunction directly corrupts prob.u0 through its aliased internal cache, will file that upstream.

ChrisRackauckas:
Show the plots I'm curious

singhharsh1708:
they're in the updated PR description too, but inline:

burgers (tsit5 mid-range, multirate below 1e-8):
burgers

ks explicit (fbdf owns it):
ks explicit

ks implicit (beats same-split kencarps 30-300x, still loses to fbdf/tsit5 unconstrained):
ks implicit

allen cahn (null case, tsit5 wins):
allen cahn

ChrisRackauckas:
Not the best result but, maybe it can keep getting tuned. It's just the start for those methods.

PR #1627, opened 2026-07-29
Fix broken LinearSolve benchmarks and refresh to LinearSolve v5

AJ0070:
Disclaimer: Work in Progress

PR #1635, opened 2026-08-08
Add SparseDirect and CacheReuse benchmarks; scaffold LinearSolveGPU folder

AJ0070:
## Summary

Follows #1634. Two new benchmark documents for `benchmarks/LinearSolve` and a new GPU folder:

**SparseDirect.jmd**: dedicated sparse direct comparison. Nine entries (UMFPACK, KLU, PureKLU, PureUMFPACK, SupernodalLU serial and threaded, Pardiso, Sparspak, and the automatic Default) on 1D/2D/3D finite difference matrices from 10 to ~40k unknowns. Reports factor+first-solve and cached re-solve separately, gates every timing on a true-residual check, and ends with a winner-per-regime table. First dedicated coverage of the new pure-Julia solvers added in #1634. ParU is excluded with an in-document explanation: repeated large ParU analyses ratchet the GC allocation accounting process-wide (SciML/LinearSolve.jl#1142, the bug behind the wedged benchmark runs on #1627).

**CacheReuse.jmd**: quantifies the init/solve! cache API. Naive re-solve vs new-b (backsolve only) vs new-A same-pattern refactorization, steady-state allocations per warm solve, and GMRES warm_start iteration savings in the time-stepping regime. Its correctness gate also surfaced that some backends (RFLU, MKL) factorize an assigned cache.A in place while others copy, now documented in the text.

**benchmarks/LinearSolveGPU/**: DenseGPUOffload.jmd (CPU vs GPU LU/QR/32-mixed offload with the PCIe round-trip measured separately and an explicit crossover N) and SparseGPU.jmd (CUDSS vs UMFPACK/KLU with the CSR H2D transfer explicit). Runner config targets the GPU runner. Both documents are validated on a Tesla T4. Note: this folder's CI job is expected to fail at instantiate for now. No Manifest can resolve until the CUDA/CUDACore/CUDSS version skew clears upstream (LinearSolve 5 requires CUDSS 0.7 which requires CUDACore >= 0.7, while current CUDA caps CUDACore at 0.6.7). The folder README documents the conflict and the unblock paths; the red check on this job is the known, documented state rather than a defect in the documents.

## Validation

Every document was validated end-to-end in isolated containers before this PR: SparseDirect (full 9-algorithm sweep) and CacheReuse (full size) pass and produce markdown; the two GPU documents were validated on a Tesla T4 at reduced sizes with all correctness gates green. The three pre-existing folder documents pass the same check on this branch's base.


ChrisRackauckas:
Also needs to handle pages.jl for the new folder.

AJ0070:
Added LinearSolveGPU to the section titles in docs/pages.jl.

ChrisRackauckas:
That edit also needs to be done on SciMLBenchmarksOutput, so copy that over. But I'll merge and we'll see if this kicks off the new machine.

AJ0070:
On it!

AJ0070:
Copied the pages.jl edit over to SciMLBenchmarksOutput: SciML/SciMLBenchmarksOutput#23

PR #1636, opened 2026-08-09
Add LinearSolveDistributed: MPI strong-scaling benchmarks for the distributed PETSc path

AJ0070:
## Summary

New folder `benchmarks/LinearSolveDistributed` benchmarking the distributed linear solver work in LinearSolve.jl (PETScAlgorithm with a communicator on replicated SparseMatrixCSC input):

**SmokeTest.jmd**: minimal 2-rank correctness gate for the harness. Runs in minutes; a nonzero worker exit fails the document.

**StrongScaling.jmd**: fixed-size 2-D FD Laplacian solved with PETSc GAMG-CG across rank counts, reporting speedup, parallel efficiency, and iteration counts. The methodology guards are the point: iteration counts print on every run (a rank-dependent preconditioner silently corrupts strong scaling into an algorithm comparison), the serial baseline is pinned to one BLAS/OpenMP thread per rank, PETSc's stopping norm is aligned with LinearSolve's true-residual safety check, and the document warns on iteration spread above 25% or efficiency above 110% instead of silently plotting an impossible curve.

**run_solve.jl / setup.sh**: the MPI worker (correctness gate on every solve, CSV output) and the runner hook that pins MPI.jl to MPICH_jll so the whole stack shares PETSc_jll's MPI ABI.

Building this harness surfaced SciML/LinearSolve.jl#1166 (solve! discards iteration count on safety-check failure) and fed into SciML/LinearSolve.jl#1142; the methodology rationale (why block-Jacobi is excluded, why the stopping norm is aligned with the safety check, why iteration counts print on every run) is documented inside the documents themselves.

## Sizes

The committed problem size and rank counts are deliberately modest so the first runner execution proves the pipeline inside a small CI budget (the full document is under an hour). A follow-up raises N toward 1e6 and extends the rank range once the harness is green on the runner; the document's closing section describes exactly that procedure.

pages.jl gets the section entry in this PR; the matching SciMLBenchmarksOutput edit is SciML/SciMLBenchmarksOutput#24.

AI Disclosure: Used Opus 5


benchmarks/LinearSolveDistributed/benchmark_config.toml
```diff
@@ -0,0 +1,8 @@
+# Runner configuration for the distributed (CPU / MPI) linear-solver benchmarks.
+#
+# CPU MPI scaling runs — stable multi-core self-hosted hardware (amdci1/amdci3).
+# The GPU benchmark (B7) lives in the sibling folder LinearSolveDistributedGPU/
+# with its own ["self-hosted","gpu","exclusive"] runner, because runner selection
+# is resolved per benchmark folder.
+runner = ["self-hosted", "benchmark"]
+timeout = 12000
```
ChrisRackauckas: why is this declared?

AJ0070: It duplicated the repository defaults so it selected nothing; copied from the GPU folders' pattern where a non-default runner genuinely needs choosing. Removed, along with two unused dependencies the same audit turned up.

AJ0070:
@ChrisRackauckas This one should be ready for another look whenever you get a chance: the config file from your review comment is removed (it duplicated the defaults), a couple of unused dependencies went with it, and all checks are green. The companion pages.jl edit for the output repo is SciML/SciMLBenchmarksOutput#24.

ChrisRackauckas:
I'm not sure this is going to work. If you mpiexec the weave locally did that work?

AJ0070:
Yes: the exact command the documents run (mpiexec with the MPICH_jll binary, 2 ranks, run_solve.jl) passes locally end to end, including on the final PR content: GAMG converges in 11 iterations with residual 5e-9 and a Success retcode, and the full weave was run during bring-up.

I also validated the same harness in isolated cloud containers (Modal) before opening this: a clean Debian image with the same MPICH_jll pin that setup.sh applies ran the identical mpiexec path at N=250k on 1/2/4/8 ranks and at N=1M on 4/8 ranks, all with correct residuals and flat GAMG iteration counts (15 to 17), so the MPI plumbing is proven outside my machine as well.

The one deliberate quirk is -launcher fork, which works around MPICH_jll hydra failing PMI bootstrap on single-node setups; it is harmless where hydra is healthy. The runner-side unknown is only setup.sh pinning MPI.jl to MPICH_jll on first run, which is why the committed sizes are small enough to fail fast if the environment disagrees.

PR #1638, opened 2026-08-10
LinearSolveGPU: commit Manifest now that CUDA 6 resolves the version conflict

AJ0070:
## Summary

The LinearSolveGPU folder merged in #1635 without a Manifest because no version set satisfied LinearSolve 5 + CUDSS 0.7 + the pre-6.0 CUDA stack (CUDACore version conflict), so its CI job failed at instantiate by construction, as documented in the folder README.

The CUDA 6 modular-stack releases resolve the conflict. This PR:

- widens the folder compat to CUDA 5-6 / CUDSS 0.6-6 and commits the resolved Manifest (CUDA 6.2.0, CUDSS 0.7.0, CUDACore 6.2.0, LinearSolve 5.9.0)
- removes the README entirely: it existed to document the blocker, no other benchmark folder carries one, and its remaining content duplicated the documents and the git history
- removes two leftover internal plan references from the document intros

## Validation

CI-equivalent rehearsal in an isolated container with a Tesla T4: `Pkg.instantiate()` on this exact committed Manifest succeeds (the step the CI job previously failed), and both documents then weave end-to-end at full size through `benchmark.jl` with all correctness gates green and markdown produced.

Merging this should turn the folder's benchmark job green and give the GPU documents their first real run on the GPU runner.


ChrisRackauckas:
It didn't trigger in the PR: let's see if it triggers on merge.

PR #1639, opened 2026-08-11
LinearSolveDistributed: add weak-scaling and serial-crossover documents

AJ0070:
## Summary

Completes the planned document set for `benchmarks/LinearSolveDistributed` (follows #1636):

**WeakScaling.jmd**: constant work per rank across rank counts, reporting weak efficiency and GAMG iteration counts with the same auditability guards as the strong-scaling document (iteration-growth and above-110% efficiency warnings). The first validation runs surfaced a structural property now documented in the text: the replicated-input path makes every rank assemble the full matrix, so assembly cost grows with total N even though per-rank solver work is constant, and measured weak efficiency reflects both effects.

**SerialCrossover.jmd**: the best serial direct factorizations (UMFPACK, KLU, SupernodalLU) against the distributed GAMG-CG path at a fixed rank count across a size sweep, answering the most actionable question about a distributed solver: when to reach for it at all. Both sides pay their honest setup costs, serial timings are correctness-gated, and the distributed iteration counts are printed so algorithmic flatness is checkable. In the committed range the serial factorizations win throughout (SupernodalLU fastest at every size), placing the crossover above it; extending the range on dedicated hardware is the documented follow-up.

Folder environment: adds Printf (used by both documents) and updates LinearSolve so SupernodalLUFactorization is available.

## Validation

All four folder documents (the two new ones and the two existing ones, which also run under the updated Manifest) were woven end-to-end in isolated containers replicating the CI path: setup.sh MPI pin, instantiate of the exact committed environment, then benchmark.jl per document. Four of four produce markdown with every correctness gate green. The two new documents were additionally validated locally at the full committed sizes.


ChrisRackauckas:
This is too much to build into the docs, it should instead be in SciMLBenchmarks with a link

AJ0070:
This one adds the documents to SciMLBenchmarks itself (benchmarks/LinearSolveDistributed, following #1636), not to any docs build; nothing here runs at docs time.

ChrisRackauckas:
Oops, I thought it was the other repo lol. Okay that makes more sense.

BTW the other is now running: https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/31533701626/job/93919680557

ChrisRackauckas:
WeakScaling_4_1
StrongScaling_5_1
StrongScaling_4_1
SerialCrossover_4_1

ChrisRackauckas:
That second plot is a bit odd? But let's merge this since it looks like it's working, and a small follow up can come.

PR #1641, opened 2026-08-12
LinearSolveDistributed: raise sizes for the runner; two-regime strong scaling

AJ0070:
## Summary

The follow-up invited in #1639. The first runner execution reproduced the known small-N cache artifact in strong scaling (9x speedup on 2 ranks, 35x on 4 at N=40k), so this PR makes that artifact part of the document's content instead of a puzzle:

**StrongScaling.jmd** now runs two sizes deliberately and plots them together: N=40k as the labeled cache-artifact regime (per-rank working set drops into cache as ranks are added, producing spectacular but meaningless superlinear speedup) and N=1e6 as the honest regime where the working set exceeds cache at every rank count. The text explains the contrast; the superlinearity warning now only flags the large series, since tripping there would mean the honest regime has not been reached.

**WeakScaling.jmd** moves to 100k unknowns per rank and extends to 8 ranks. **SerialCrossover.jmd** extends the sweep to 1e6 unknowns to bracket the crossover the first run placed above 90k. Both scaling documents now cover ranks 1 through 8.

## Budget

The runner completed the whole folder in 20 minutes at the previous sizes. Scaling the measured per-solve times to the raised sizes puts the folder around two hours, well inside limits; the N=1e6 rank-1 baseline in StrongScaling is the dominant term.

## Validation

WeakScaling and SerialCrossover were woven end-to-end at the raised sizes in isolated containers replicating the CI path (setup.sh MPI pin, exact committed environment, benchmark.jl per document): both pass with markdown produced and all correctness gates green. StrongScaling's container run exceeded the container time budget on slower cloud hosts (its N=1e6 rank-1 baseline is a long serial solve there); its mechanics are identical to the already runner-proven version, its restructured two-series logic was verified locally, and its N=1e6 distributed path is exercised by SerialCrossover's passing run, so it is submitted on that evidence plus the budget arithmetic above.

### DIFF
--- a/benchmarks/LinearSolveDistributed/SerialCrossover.jmd
+++ b/benchmarks/LinearSolveDistributed/SerialCrossover.jmd
@@ -67,9 +67,8 @@ serial_algs = [
     ("SupernodalLU", SupernodalLUFactorization()),
 ]
 
-# Target unknown counts (the worker rounds to a square grid). Modest for CI;
-# the closing section describes raising the range on dedicated hardware.
-const SIZES = [10_000, 22_500, 40_000, 90_000]
+# Target unknown counts (the worker rounds to a square grid).
+const SIZES = [40_000, 90_000, 250_000, 1_000_000]
 ```
 
 ## Methodology
--- a/benchmarks/LinearSolveDistributed/StrongScaling.jmd
+++ b/benchmarks/LinearSolveDistributed/StrongScaling.jmd
@@ -57,9 +57,13 @@ end
 
 ## Run the rank sweep
 
-`N` here is the target number of unknowns. It is deliberately modest so the notebook
-runs in CI in minutes; on the benchmark hardware this is raised to expose scaling on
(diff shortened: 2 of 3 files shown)


AJ0070:
The three benchmark jobs here were cancelled by the 24 hour queue limit without starting. It looks like the runner has been occupied by two runs since Aug 10 and Aug 11, and everything queued behind them is being auto cancelled at the 24 hour mark, including the master runs from yesterday. Could those two runs be checked and the jobs here retriggered when a slot frees up?

AJ0070:
@ChrisRackauckas The benchmark jobs here have now been cancelled three times by the 24 hour queue limit; the runner's slots have been held for days by long shards, so a PR run cannot get through at the moment. The documents are runner-proven at the previous sizes and two of the three were woven end-to-end at the raised sizes in containers. Happy to leave it queued, but if you are comfortable merging on that evidence, the master rebuild will produce the pages when the runner frees up.

PR #1655, opened 2026-08-15
Dense GPU offload: compare against the best CPU baseline, not just OpenBLAS

AJ0070:
## Summary

Follow-up to the question on SciMLBenchmarksOutput#25: the published DenseGPUOffload page computed its crossover against plain OpenBLAS `LUFactorization`, which is not the CPU performance a LinearSolve user actually gets on the runner. A crossover against a slow baseline is not a crossover.

The CPU side is now the best of four options, all through the same `LinearProblem` API: `LUFactorization` (OpenBLAS), `MKLLUFactorization`, `RFLUFactorization`, and the LinearSolve default algorithm choice, which is the number a user gets by not specifying an algorithm at all. The document now prints the CPU model and thread counts alongside the GPU name, tables every algorithm, prints a per-size best-CPU versus best-GPU line naming which CPU algorithm set the bar, and computes the crossover against the best CPU option.

The folder environment gains RecursiveFactorization and MKL_jll, following the conventions of benchmarks/LinearSolve. The Manifest update is purely additive: CUDA 6.2.0 and CUDSS 0.7.0 stay exactly where the runner has them.

## Validation

Both documents in the folder were woven end-to-end on a T4 in a container replicating the CI path (exact committed Project and Manifest, benchmark.jl per document): both pass with markdown produced and every correctness gate green. On that host the crossover moves from N = 256 against OpenBLAS alone to N = 512 against the best CPU, with MKL LU setting the CPU bar at every size and GPU LU reaching roughly 11x over it by N = 8192. The strengthened baseline makes the GPU claim smaller and true rather than large and hollow; the runner's V100 will publish its own numbers.

AI Disclosure: Used Opus 5


AJ0070:
The first runner execution surfaced that LinearSolve deliberately does not load MKL_jll on AMD EPYC (LinearSolve.jl#518, the LoadMKL_JLL preference), and the GPU runner is an EPYC 9354, so the MKL column came back NaN with load warnings. The document now probes MKL once through the public API and includes that row only where LinearSolve actually loads it, printing the policy note otherwise. On the runner the CPU bar is set by RFLU throughout, and the crossover against the best CPU option stays at N = 512. Note the same unguarded MKLLUFactorization exists in benchmarks/LinearSolve/LUFactorization.jmd where it will error the whole sweep on the next master rebuild; a separate PR for that is prepared.

PR #1659, opened 2026-08-15
LUFactorization: probe MKL availability instead of assuming it

AJ0070:
## Summary

Same issue #1655 fixed for the GPU folder, but here it is page-killing: LinearSolve deliberately does not load MKL_jll on AMD EPYC (LinearSolve.jl#518, the LoadMKL_JLL preference), and the benchmark runner is an EPYC 7502. Under LinearSolve v5 the unguarded `MKLLUFactorization()` entry in this document throws on first use, which errors the whole sweep chunk and every plot below it.

The currently published page does not show this only because it predates the v5 refresh: its appendix pins LinearSolve v2.5.0, which is older than the EPYC gate. The break surfaces on the next master rebuild of benchmarks/LinearSolve.

The fix probes MKL once through the public API and includes it in the algorithm list only where LinearSolve actually loads it, printing the exclusion note on the page otherwise. No other document in the folder is exposed: CacheReuse already guards its solves, and the Pardiso entries load MKL through Pardiso.jl, which has no EPYC gate.

## Validation

The full document was woven end-to-end in a container replicating the CI path on an MKL-enabled host: no errors, markdown produced, all seven algorithms benchmarked. The exclusion branch was exercised separately by forcing the probe false and running the sweep logic: six algorithms, all rows filled, note printed. The runner execution of the analogous probe in #1655 confirmed the EPYC branch behaves as designed on the machines that publish.

AI Disclosure: Used Opus 5

### DIFF
--- a/benchmarks/LinearSolve/LUFactorization.jmd
+++ b/benchmarks/LinearSolve/LUFactorization.jmd
@@ -24,19 +24,21 @@ function luflop(m, n = m; innerflop = 2)
     end
 end
 
-# LinearSolve deliberately does not load MKL_jll on CPUs it defaults away from,
-# notably AMD EPYC, which is this benchmark runner's CPU (LinearSolve.jl#518,
-# the LoadMKL_JLL preference). On such machines MKLLUFactorization throws on
-# first use, which would kill this whole sweep; probe once and include it only
-# where it runs, so the page states the exclusion instead of erroring.
+# LinearSolve's LoadMKL_JLL preference defaults to false on AMD EPYC (this
+# runner's CPU; SciML/LinearSolve.jl#518), and when it is false MKL_jll is not loaded
+# at all, so MKLLUFactorization throws instead of running. This folder's
+# Project.toml sets the preference to true precisely so MKL IS measured here:
+# whether MKL still loses to OpenBLAS and RF on EPYC is a question this
+# benchmark should answer, not assume. The probe is a safety net: if MKL
+# still cannot load, say so on the page rather than erroring the sweep.
 mkl_works = try
     solve(LinearProblem(Matrix(8.0I, 8, 8), ones(8)), MKLLUFactorization())
     true
 catch
     false
 end
-mkl_works || println("MKLLUFactorization excluded: LinearSolve does not load ",
-    "MKL_jll on this CPU by default (LoadMKL_JLL preference; LinearSolve.jl#518).")
+mkl_works || println("MKLLUFactorization excluded: MKL failed to load in this ",
+    "environment even with LoadMKL_JLL = true (see SciML/LinearSolve.jl#518).")
 
 algs = [LUFactorization(), GenericLUFactorization(), RFLUFactorization(),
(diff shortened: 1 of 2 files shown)


benchmarks/LinearSolve/LUFactorization.jmd
```diff
@@ -24,8 +24,23 @@ function luflop(m, n = m; innerflop = 2)
...
+# first use, which would kill this whole sweep; probe once and include it only
+# where it runs, so the page states the exclusion instead of erroring.
+mkl_works = try
+    solve(LinearProblem(Matrix(8.0I, 8, 8), ones(8)), MKLLUFactorization())
+    true
+catch
+    false
+end
+mkl_works || println("MKLLUFactorization excluded: LinearSolve does not load ",
+    "MKL_jll on this CPU by default (LoadMKL_JLL preference; LinearSolve.jl#518).")
```
ChrisRackauckas: Why not use the mkl isavailable function?

AJ0070: MKL_jll.is_available() returns true on the EPYC runner: the binary is installed and loadable, but LinearSolve still refuses it through the LoadMKL_JLL preference, so gating on is_available would include MKL exactly where MKLLUFactorization throws. The decision LinearSolve actually makes is the internal LinearSolve.usemkl, which combines the preference, is_available, and a version floor. The probe solve tests the exact contract the sweep depends on and stays correct if that gating logic ever changes. Happy to switch to LinearSolve.usemkl if you prefer the direct check.

ChrisRackauckas: That's weird, MKL should work on this computer, is the threshold correct?

AJ0070: MKL itself works fine on it. What blocks it is LinearSolve's LoadMKL_JLL preference, whose default is `!occursin("EPYC", cpubrand())` from SciML/LinearSolve.jl#518, and when that is false LinearSolve does not load MKL_jll at all, so an explicit MKLLUFactorization() throws "MKL binary is missing" rather than just not being the default. That is the exact error in the runner log. Excluding MKL was the wrong fix for a benchmark: measuring MKL on EPYC is exactly what tells whether the SciML/LinearSolve.jl#518 default still holds. I will set LoadMKL_JLL = true for this folder in Project.toml preferences, the same way LinearSolve's own docs build does in docs/LocalPreferences.toml, keep the probe only as a safety net that prints a note if loading fails, and do the same for the GPU folder in a follow-up. Separately, an explicit MKLLUFactorization request probably should load MKL regardless of the default-choice preference; happy to open that on LinearSolve.jl if you agree.

AJ0070:
Pushed the preference-based fix: benchmarks/LinearSolve/Project.toml now sets LoadMKL_JLL = true for LinearSolve, so MKL loads on the EPYC runner and MKLLUFactorization is measured rather than skipped. The probe stays only as a safety net. Woven end-to-end in a container with the preference active: seven algorithms, no errors. The runner execution here will be the first real MKL-on-EPYC numbers for this page.

PR #1674, opened 2026-08-24
Refresh 12 benchmark Manifests to pick up OrdinaryDiffEqDifferentiation 3.10

ChrisRackauckas:
Follow-up to #1670. **Please read the CI cost note before merging.**

## Why

DAE's `fekete.jmd` failed in CI ([run 32662738361](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/32662738361)) with `No matching function wrapper was found!`. `get_fresh_jacobian` evaluated a user-supplied **in-place** Jacobian **out-of-place** as `f.jac(u, p, t)`, hitting a `FunctionWrappersWrapper` that only carries the in-place signature. The path is reached whenever an implicit solver aborts and `check_error` calls `log_numerical_instability` — which `verbose` cannot suppress. So it is latent in **any** folder where a problem supplies an in-place `jac` and a solver fails at tight tolerance, i.e. exactly what a work-precision diagram does on purpose.

`OrdinaryDiffEqDifferentiation` **3.9.0** fixed it (`get_fresh_jacobian` branches on `SciMLBase.isinplace` and calls `calc_J!`); **3.10.0** refines the branch for a non-`AbstractMatrix` `J`. Every benchmark Manifest in the repo pinned **below 3.9.0**.

## What this does

A targeted `Pkg.update("OrdinaryDiffEqDifferentiation")` does **not** move it — the OrdinaryDiffEq sub-packages that depend on it (Rosenbrock, BDF, SDIRK, FIRK, …) have to move first. So these are full `Pkg.update()` refreshes, each resolved with the Julia version named in that folder's own Manifest (all `1.11.9` here, so **no CI Julia-version change**). SciMLBase goes 3.39.1 → 3.49.2 in each.

| folder | before | after |
|---|---|---|
| AdaptiveSDE | 3.0.0 | 3.10.0 |
| AutomaticDifferentiation | 3.7.0 | 3.10.0 |
| DynamicalODE | 3.1.1 | 3.10.0 |
| ModelingToolkit | 3.0.0 | 3.10.0 |
| MultiLanguage | 3.0.0 | 3.10.0 |
| NBodySimulator | 3.0.0 | 3.10.0 |
| NonStiffBVP | 3.1.1 | 3.10.0 |
| NonStiffDDE | 3.1.1 | 3.10.0 |
| NonStiffODE | 3.1.1 | 3.10.0 |
| NonStiffSDE | 3.0.0 | 3.10.0 |
[...]

PR #1669, opened 2026-08-23
StiffODE: make the folder finish again (lsoda() hang, HeatingSystem blowup, #1642 ParseError)

ChrisRackauckas:
## What changed, and when did `benchmarks/StiffODE` last work

**Last complete build of the folder: source commit `0c46d59a` (2026-03-14), published to
SciMLBenchmarksOutput on 2026-03-18T08:31Z.** That build produced all 8 markdown files the
folder had at the time (`Bruss`, `E5`, `Hires`, `MKMBatteryChemistry_wpd`, `Orego`,
`Pollution`, `ROBER`, `VanDerPol`). It ran on the old Buildkite pipeline — the
Buildkite→GHA migration landed later the same day (2026-03-18T11:52Z) — so there is no GHA
job duration for it.

**Between 2026-03-13 and 2026-03-17 the folder went from 8 `.jmd` files to 13:**

| commit | date | files added |
|---|---|---|
| `32796cb` "Add 4 stiff ODE benchmarks from IVP Test Set Fortran translations" | 2026-03-13 | `Beam.jmd`, `EMEP.jmd`, `MedAkzoNobel.jmd`, `RingModulator.jmd` |
| `ecc8aa4` "Add heating system benchmark for stiff ODE solvers" | 2026-03-16 | `HeatingSystem_wpd.jmd` |

**None of those five files has ever appeared in SciMLBenchmarksOutput.** `markdown/StiffODE`
has been touched exactly once since March — 2026-07-25, and that commit changed only
`Bruss.md` (a per-file run triggered by a `Bruss.jmd`-only change; those take 1.2–1.3 h and
succeed). So the whole-folder job has not finished since 2026-03-18, and the reason is
the five files added right after it.

Whole-folder CI attempts since:

* run `31675055195` ("Add NordsieckBDF to StiffODE benchmarks", 2026-08-13) — cancelled after 24 h
* the 2026-08-14 `amdci3-1` run — cancelled at **119.9 h**, still not done

Manifest drift over the same window (for context — none of it is the primary cause):
julia `1.10.10 → 1.11.9`, `LSODA 0.7.5 → 1.1.0`, `OrdinaryDiffEq 6.108 → 7.2.1`,
`ODEInterfaceDiffEq 3.15 → 4.1`, `Sundials 5.1 → 6.4.2`, `DiffEqDevTools 2.49 → 3.2`.

## Measured root causes

All numbers below are wall clock on an Apple M-series laptop (12 cores). "Probe" = a
[...]

PR #1676, opened 2026-08-24
StiffODE: relax LinearSolve/AlgebraicMultigrid caps to reach OrdinaryDiffEqDifferentiation 3.10

ChrisRackauckas:
## Why

`benchmarks/StiffODE` was stuck on `OrdinaryDiffEqDifferentiation v3.4.1`, which
contains a real bug that this folder hits on essentially every work-precision
diagram: `get_fresh_jacobian` evaluates a user-supplied **in-place** Jacobian
out-of-place as `f.jac(u, p, t)`. For an MTK-generated `jac = true` problem that
lands on a `FunctionWrappersWrapper` carrying only the in-place signature, so it
throws

```
ERROR: No matching function wrapper was found!
```

The path is reached whenever an implicit solver aborts and `check_error` calls
`log_numerical_instability` — which `verbose` cannot suppress — i.e. exactly what
happens at the tight tolerances a WPD sweeps. It was fixed in
`OrdinaryDiffEqDifferentiation` 3.9.0 (branch on `SciMLBase.isinplace`, call
`calc_J!`) and refined in 3.10.0.

#1674 refreshed 12 folders that could move; StiffODE could not — a full
`Pkg.update()` left `OrdinaryDiffEqDifferentiation` at 3.4.1 and `SciMLBase` at
3.39.1. This PR unblocks it.

## What was actually capping it

Two `[compat]` entries in `benchmarks/StiffODE/Project.toml`, found by asking the
resolver directly (`Pkg.add(PackageSpec(name = "OrdinaryDiffEqDifferentiation",
version = "3.9.0"))` on Julia 1.11.9, the folder Manifest's `julia_version`):

1. **`LinearSolve = "3"`** — the primary cap.
   ```
   LinearSolve [7ed4a6bd] log:
    ├─restricted to versions 3 by project, leaving only versions: 3.0.0 - 3.87.0
    └─restricted by compatibility requirements with OrdinaryDiffEqDifferentiation
      to versions: 5.8.0 - 5.13.0 — no versions left
   ```

2. **`AlgebraicMultigrid = "1"`** — exposed once (1) was relaxed. AMG 1.x caps
   LinearSolve below 4:
   ```
   LinearSolve [7ed4a6bd] log:
    ├─restricted by compatibility requirements with OrdinaryDiffEqDifferentiation
    │ to versions: 5.8.0 - 5.13.0
    └─restricted by compatibility requirements with AlgebraicMultigrid
      to versions: 2.0.0 - 3.87.0 — no versions left
   ```

**Sundials was a victim, not a cause.** Its own `[compat] = "6"` already admitted
6.6.0; it sat at 6.4.2 transitively because the LinearSolve cap pinned SciMLBase
at 3.39.1. Nothing about the Sundials bound needed changing, and it moved on its
own once the two entries above were relaxed.

**`MethodOfLines = "0.11"` and `GeometricIntegratorsDiffEq = "1"` do NOT block
this chain.** `Pkg.status(; outdated = true, mode = PKGMODE_MANIFEST)` reports
them as outdated-and-capped, but with the two relaxations above the environment
resolves to `OrdinaryDiffEqDifferentiation v3.10.0` with both still pinned where
they are. They are left alone here — see "Left for follow-up".

## Changes

`benchmarks/StiffODE/Project.toml`:

| entry | before | after |
|---|---|---|
| `LinearSolve` | `"3"` | `"3, 4, 5"` |
| `AlgebraicMultigrid` | `"1"` | `"1, 2"` |
| `GenericSchur` | (transitive) | added to `[deps]`, `"=0.5.6"` — temporary, see below |

plus the regenerated `Manifest.toml`.

### The GenericSchur pin (temporary workaround for an upstream regression)

Moving to LinearSolve 5 lets `ExponentialUtilities` advance 1.31.0 → 1.35.0
(1.35.0 declares `LinearSolve = "4.1, 5"`, which is why it was previously held
back). **ExponentialUtilities 1.35.0 cannot precompile at all against
GenericSchur 0.5.7**, and `OrdinaryDiffEqExponentialRK` — a direct dep of this
folder — depends on it:

```
ERROR: LoadError: MethodError: no method matching geigen!(::SymTridiagonal{Float64, SubArray{...}}, ::QRIteration; sortby::typeof(identity))
  [3] expT!  @ ExponentialUtilities/src/krylov_phiv_error_estimate.jl:62
  [4] expv!  @ ExponentialUtilities/src/krylov_phiv_error_estimate.jl:191
 [22] macro expansion @ PrecompileTools/src/workloads.jl:140
```

The regression is in GenericSchur 0.5.7. Its new `pirates.jl` defines

```julia
function LinearAlgebra.eigen!(A::SymTridiagonal{<:AbstractFloat};
        alg::Algorithm = QRIteration(), sortby::Union{Function, Nothing} = nothing)
    return geigen!(A, alg; sortby = (sortby === nothing ? identity : sortby))
end
```

but `symtridiag.jl` defines `geigen!(A::SymTridiagonal{T}, alg = QRIteration())`
with **no keyword arguments at all** — so *every* call to `eigen!` on a real
`SymTridiagonal` under 0.5.7 is a `MethodError`. 0.5.6 forwarded `kwargs...`
(empty by default) and works.

Reproduced in a bare two-package environment, independent of this repo:

| combination | result |
|---|---|
| ExponentialUtilities 1.35.0 + GenericSchur 0.5.7 | ✗ precompile fails |
| ExponentialUtilities 1.35.0 + GenericSchur 0.5.6 | ✓ |
| ExponentialUtilities 1.31.0 + GenericSchur 0.5.7 | ✓ |

GenericSchur's `piracy = "false"` preference is *not* a workaround — with piracy
off there is no `eigen!(::SymTridiagonal)` method at all for the non-strided
views `ExponentialUtilities` builds, so it fails the same way. 0.5.7 is the
latest registered version, so the only option available inside this repo is to
hold GenericSchur at 0.5.6 (ExponentialUtilities 1.35.0 declares
`GenericSchur = "0.5.3"`, so 0.5.6 is admissible). **Remove this entry once
GenericSchur is fixed upstream.** Worth reporting to GenericSchur.jl; other
folders that reach LinearSolve 5 while depending on `OrdinaryDiffEqExponentialRK`
will hit the same wall.

## API-breakage audit

Every major version crossed was checked against the 13 `.jmd` files in the folder.

**LinearSolve 3 → 5** (crosses 4 and 5). The breaking changes are batched-RHS
`LinearProblem(A, B)` (v4) and `LinearSolution.cache` always being `nothing` plus
removal of `cleanup_*_cache!` (v5). All are unreachable here: grepping the folder
for `LinearProblem`, `solve!`, `.cache`, `cleanup_`, `LinearSolve.`, `init(`
returns **nothing**. The folder only uses algorithm constructors
(`KLUFactorization`, `UMFPACKFactorization`, `RFLUFactorization`,
`KrylovJL_GMRES`) and the `precs` builder — and `precs` is still
`(A, p) -> (Pl, Pr)` in 5.13.0, unchanged from 3.87.0. All four constructors are
still exported. No `.jmd` edits needed.

**AlgebraicMultigrid 1 → 2**. `Bruss.jmd` uses `ruge_stuben(W; presmoother, postsmoother)`,
`AlgebraicMultigrid.Jacobi(x)` and `aspreconditioner(ml)` (lines 302, 336-338,
351-353). In 2.0.1 `ruge_stuben` still takes `presmoother`/`postsmoother` keywords
(it gained a `symmetry` keyword and a `_default_coarse_solver`, neither of which
this code passes), `Jacobi(x::AbstractArray, ω = 0.5)` still exists, and
`aspreconditioner` is unchanged. Verified by running the exact `Bruss.jmd`
preconditioner code (see below). No `.jmd` edits needed.

**MethodOfLines / GeometricIntegratorsDiffEq**: not crossed — left at their
current pins, so no migration was required.

## Final versions

| package | before | after |
|---|---|---|
| OrdinaryDiffEqDifferentiation | 3.4.1 | **3.10.0** |
| SciMLBase | 3.39.1 | **3.49.2** |
| Sundials | 6.4.2 | **6.6.0** |
| LinearSolve | 3.87.0 | **5.13.0** |
| AlgebraicMultigrid | 1.2.0 | **2.0.1** |
| OrdinaryDiffEq | 7.2.1 | 7.7.0 |
| OrdinaryDiffEqRosenbrock | 2.4.2 | 2.7.0 |
| OrdinaryDiffEqNonlinearSolve | 2.0.2 | 2.9.0 |
| ModelingToolkit | 11.38.2 | 11.39.1 |
| NonlinearSolve | 4.21.0 | 4.28.0 |
| GenericSchur | 0.5.6 | 0.5.6 (held) |
| MethodOfLines | 0.11.19 | 0.11.19 (unchanged) |
| GeometricIntegratorsDiffEq | 1.3.3 | 1.3.3 (unchanged) |

`julia_version` in the Manifest stays `1.11.9`.

## Verified locally

macOS aarch64, Julia 1.11.9, `--project=benchmarks/StiffODE`.

- The environment resolves, instantiates, and precompiles. `Pkg.status(; outdated
  = true, mode = PKGMODE_MANIFEST)` no longer lists
  `OrdinaryDiffEqDifferentiation`, `SciMLBase`, `Sundials`, `LinearSolve` or
  `AlgebraicMultigrid` — they are all at the newest admissible version.
- Every package the 13 `.jmd` files load precompiles and loads, **including
  `ODEInterfaceDiffEq`** (it built fine on this machine).
- A fresh `Pkg.update()` on the committed `Project.toml` produces no changes —
  the `GenericSchur = "=0.5.6"` pin holds and nothing else wants to move.
- Smoke script A (4 min 19 s wall on a warm depot) — reduced Brusselator (N = 8 instead of N = 32) exercising the
  API surface the relaxations touch, all passing:
  - `Symbolics.jacobian_sparsity` + `modelingtoolkitize(prob; jac = true, sparse = true)`
  - `KenCarp47(linsolve = KrylovJL_GMRES(; precs = incompletelu), concrete_jac = true)`
    and the same with `precs = algebraicmultigrid` — the exact `precs` closures
    from `Bruss.jmd`
  - `KenCarp47` with `KLUFactorization`, `UMFPACKFactorization`, `KrylovJL_GMRES`
  - the `Bruss.jmd` Sundials preconditioner block verbatim: the `ilu`-based
    `psetupilu`/`precilu` and the AMG-based
    `aspreconditioner(ruge_stuben(W, presmoother = AlgebraicMultigrid.Jacobi(...), postsmoother = ...))`
    `psetupamg`/`precamg`, driven through
    `CVODE_BDF(linear_solver = :GMRES, prec = ..., psetup = ..., prec_side = 1)`
  - `CVODE_BDF(linear_solver = :KLU)` on the MTK problem
  - a `WorkPrecisionSet` at `abstol = 1e-8 … 1e-11` on the MTK problem — the code
    path that used to throw `No matching function wrapper was found!`
  - `MOLFiniteDifference` + `discretize` on the Brusselator PDE system, solved
    with `CVODE_BDF(linear_solver = :KLU)`
- Smoke script B — ROBER with a dense analytic Jacobian (the shape
  `Pollution.jmd` / `HeatingSystem_wpd.jmd` / `MKMBatteryChemistry_wpd.jmd`
  actually use), all passing:
  - `ImplicitEulerExtrapolation`, `ImplicitEulerBarycentricExtrapolation` and
    `ImplicitHairerWannerExtrapolation`, each with `linsolve = RFLUFactorization()`
  - a `WorkPrecisionSet` down to `abstol = 1e-12` over `Rodas5P`, `KenCarp4`,
    `FBDF`, `RadauIIA5`, `CVODE_BDF` and `ImplicitEulerExtrapolation`

## A pre-existing bug this surfaced (NOT introduced here, NOT fixed here)

While writing the smoke scripts I hit, in `MedAkzoNobel.jmd`'s shape — an
`ODEFunction` with **both** `jac` and a non-`nothing` `jac_prototype` — solved
with an extrapolation method:

```
MethodError: no method matching SciMLOperators.WOperator(::ODEFunction{...}, ::Float64, ::Bool)
  [1] alg_cache(alg::ImplicitEulerExtrapolation{...})
      @ OrdinaryDiffEqExtrapolation/src/extrapolation_caches.jl:260
```

`OrdinaryDiffEqExtrapolation` calls `WOperator(f, dt, true)` in the
`has_jac(f) && f.jac_prototype !== nothing` branch. This is **not** a regression
from this PR:

- the call site is byte-identical in 2.3.1 (the version on master) and 2.5.0 —
  the only diff in that function is `SciMLBase.has_Wfact` → `_has_Wfact`;
- `WOperator` resolves to `SciMLOperators.WOperator` under both resolves
  (`OrdinaryDiffEqDifferentiation` 3.x defines no `WOperator` of its own), and
  neither SciMLOperators 1.26.1 (master) nor 1.29.0 (here) defines any
  three-positional-argument constructor — only
  `WOperator{IIP}(mass_matrix, gamma, J, u, jacvec)`.

So `MedAkzoNobel.jmd`'s extrapolation chunks were already producing this error on
master; it is invisible in CI because `weave_file` sets
`Weave.set_chunk_defaults!(:error => false)`, so the error is captured into the
rendered markdown instead of failing the build. That belongs with the runtime
fixes in #1669, not in a dependency PR.

Related observation, same category: a plain `Pkg.update()` on **master's**
unmodified `benchmarks/StiffODE/Project.toml` lands on `OrdinaryDiffEqBDF` 2.4.4
against `SciMLBase` 3.39.1, which fails to precompile with
`UndefVarError: AutoDespecialize not defined in SciMLBase`. The bounds on master
were not merely stale, they were no longer self-consistent.

## NOT verified locally

- **No `.jmd` was woven end to end.** The folder is a multi-hour job (#1669 is
  open for exactly that reason); weaving `Bruss.jmd` alone at its real `N = 32`
  is not a laptop task. Full weaving is left to CI.
- Benchmark *numbers* — the curves will be regenerated by CI.
- Only macOS aarch64 was exercised; the Linux self-hosted runners are untested here.
- The `MedAkzoNobel.jmd` `WOperator` failure above was diagnosed from the package
  sources plus a reproduction on the new resolve; I did **not** run it on
  master's committed Manifest (that environment does not precompile, see above).

## Relationship to other open work

- **#1669** (StiffODE runtime fixes) is open on this same folder. This PR is
  **dependency-only**: no `.jmd` file is touched, only `Project.toml` and
  `Manifest.toml`. It should be trivially rebasable either way round.
- Same root cause as the LinearSolve cap found in `benchmarks/Bio`; StiffODE
  additionally needed the `AlgebraicMultigrid` bound because it is one of the few
  folders that uses AMG directly.

## Left for follow-up (deliberately not done here)

- **`GeometricIntegratorsDiffEq = "1"` is dead weight.** Every single reference to
  it in this folder is commented out — `#using GeometricIntegratorsDiffEq` /
  `#sol = solve(prob, GIRadIIA3(), ...)` in `Hires.jmd:60`, `Orego.jmd:47`,
  `Pollution.jmd:229`, `VanDerPol.jmd:49`, `ROBER.jmd:44`, and nothing else. It
  is not `using`-ed by any live chunk. Keeping it pinned at 1.3.3 drags
  `GeometricIntegrators 0.16`, `GeometricIntegratorsBase`, `SimpleSolvers`,
  `RungeKutta`, `QuadratureRules` and `CompactBasisFunctions` into every
  instantiate of this folder for zero benefit. **Dropping the dependency
  outright** (rather than migrating to 2.x) would be the clean fix; it is left
  out of this PR to keep the diff dependency-only and reviewable.
- **`MethodOfLines = "0.11"` → `"1"`** is a genuine major-version migration and is
  used for real in `Bruss.jmd` (`MOLFiniteDifference`, `discretize`,
  `wrap = Val(false)`). `benchmarks/MethodOfLinesPDE` carries the identical pin,
  so that migration belongs in one PR covering both folders.
- `DomainSets = "0.7"`, `ODEInterfaceDiffEq = "4"`, `TimerOutputs = "0.5"` remain
  outdated-and-capped; none of them block the SciMLBase chain.
- The `GenericSchur = "=0.5.6"` pin should be removed as soon as GenericSchur
  ships a fix.

https://claude.ai/code/session_01KrTtUuDkfgNEq3asQuZ3F6


ChrisRackauckas-Claude:
Confirmed: GenericSchur 0.5.8 fixes the regression tracked at https://github.com/RalphAS/GenericSchur.jl/issues/24. In clean Julia 1.11 environments, GenericSchur 0.5.7 plus ExponentialUtilities 1.35.0 reproduced the `geigen!` MethodError, while 0.5.8 loaded successfully. I opened the focused stacked draft https://github.com/SciML/SciMLBenchmarks.jl/pull/1693 to remove the direct dependency and exact pin; the resulting StiffODE environment resolves unchanged, precompiles, and loads OrdinaryDiffEqExponentialRK.

PR #1801, opened 2026-09-01
Use isolated Conda Python and R for MultiLanguage benchmarks

ChrisRackauckas-Claude:
> [!IMPORTANT]
> Ignore this PR until reviewed by @ChrisRackauckas.

## What changed and why

The MultiLanguage runner previously built PyCall against `/usr/bin/python3`, whose environment does not contain SciPy, and RCall had no managed R configuration. This adds the benchmark setup hook that selects Conda for both wrappers, places them in a dedicated persistent MultiLanguage Conda root, and loads Conda's C++ runtime before Julia's older bundled copy so the installed `libR` can be opened.

The dedicated root is important on the persistent runners: resolving `r-base` against the existing shared 243-package Julia Conda environment remained CPU-bound in the classic solver for the full one-hour local bound. A clean benchmark-specific root installed a current Miniforge environment and resolved R without inheriting unrelated constraints.

The hook is scoped to MultiLanguage and preserves an explicitly supplied `CONDA_JL_HOME`.

## Failing before

The previous full runner failed before any benchmark comparison ran:

```text
ModuleNotFoundError: No module named 'scipy'
```

https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/33042289225/job/98418417866

The setup regression test failed on unmodified master because the hook did not exist. After the initial shared-root setup, the same exact test failed when changed to require the isolated fallback:

```text
Test Summary:                          | Pass  Fail  Total
Core                                   |   85     1     86
  MultiLanguage managed language setup |          1      1

Evaluated: .../.julia/conda/3/x86_64 == .../.julia/conda/SciMLBenchmarks/MultiLanguage
```

The real shared-root R build was bounded externally with Bash and timed out during dependency solving:

```text
Collecting package metadata (repodata.json): done
Solving environment: working...
# exit 124 after 3600 seconds
```

## Passing after

The same repository regression test passes with the dedicated fallback:

```text
Test Summary: | Pass  Total   Time
Core          |   86     86  37.6s
Testing SciMLBenchmarks tests passed
```

I then built Conda and RCall against an empty isolated root while applying the exact `PYTHON`, `R_HOME`, `CONDA_JL_HOME`, and `LD_LIBRARY_PATH` setup. The install completed and a real R evaluation passed:

```text
ISOLATED_RCALL_PASS result=2.0 Rhome=.../.validation/conda-multilanguage/lib/R
```

[...]

PR #1810, opened 2026-09-04
Rebuild benchmarks after support-file changes

ChrisRackauckas-Claude:
## Review status

Ignore this PR until it has been reviewed by @ChrisRackauckas.

## What changed and why

The benchmark detector previously scheduled individual `.jmd` changes and full-folder rebuilds for `.toml` changes, but silently ignored every other file below a benchmark project. That caused https://github.com/SciML/SciMLBenchmarks.jl/pull/1801 to skip its benchmark even though it changes `benchmarks/MultiLanguage/setup.sh`.

This treats every non-`.jmd` file below a benchmark project as a full-folder trigger. It covers setup scripts, auxiliary source/data files, and `benchmark_config.toml`; changed notebooks remain individually targeted, and the existing project coalescing still suppresses redundant notebook jobs.

## Failing before / passing after

I added the same setup-file fixture and ran the same detector test before and after the implementation change.

Before:

```text
Diffing against push event base SHA: e8b9a88ea051c0c63cb4121f20e625d50d28e57f
No benchmark changes detected.
Expected benchmarks/Alpha for a support-file change, got:

Process exited 1
```

After:

```text
Diffing against push event base SHA: cd39823dfa78c121a7418ad47a9114f1f6bd293f
Detected benchmark targets:
  benchmarks/Alpha -> runner=["self-hosted", "benchmark"] timeout=12000 julia=1.11
Push detection selected only changes since the event's before SHA.
Process exited 0
```

## Local verification

```text
$ bash .github/scripts/test-detect-changed-benchmarks.sh
Process exited 0

$ GROUP=Core julia +1.11.9 --project=.validation/root-env -e 'using Pkg; Pkg.test("SciMLBenchmarks")'
Core | 85 passed / 85 total | 57.9s

$ GROUP=QA julia +1.11.9 --project=.validation/root-env -e 'using Pkg; Pkg.test("SciMLBenchmarks")'
QA | 21 passed / 21 total | 2m04.2s
```

Also passed `bash -n` for both changed scripts, `typos` over both changed scripts, and `git diff --check`. Runic is not applicable to this shell-only diff. `shfmt` is not installed on this host, and this repository does not declare it as a formatter. No documentation or public API changed.

https://chatgpt.com/codex/tasks/01a03a17-ad6f-7131-82fc-d0fd57ea6512

### DIFF
--- a/.github/scripts/detect-changed-benchmarks.sh
+++ b/.github/scripts/detect-changed-benchmarks.sh
@@ -4,7 +4,7 @@ set -eo pipefail
 # Detect changed benchmark files and produce a JSON matrix for GHA.
 # Replicates the project-coalescing logic:
 #   - A changed .jmd file triggers a rebuild of just that file
-#   - A changed .toml file (Project.toml/Manifest.toml) triggers a rebuild of its entire benchmark directory
+#   - Any other changed file triggers a rebuild of its entire benchmark directory
 #   - If a directory is already being rebuilt, individual .jmd files in it are suppressed
 #
 # Reads runner configuration from:
@@ -60,8 +60,6 @@ find_project() {
 
 while IFS= read -r f; do
     [[ -z "${f}" ]] && continue
-    # Skip benchmark_config.toml changes — they don't trigger rebuilds
-    [[ "${f}" == */benchmark_config.toml ]] && continue
     proj=$(find_project "$(dirname "${f}")")
     if [[ -z "${proj}" ]]; then
         echo "::warning::Unable to find project for ${f}"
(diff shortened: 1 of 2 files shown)


ChrisRackauckas-Claude:
A fresh real-world confirmation of this detector gap is https://github.com/SciML/SciMLBenchmarks.jl/pull/1823. It changes `benchmarks/NonStiffODE/enright_pryce.jl` and its focused regression test, but the benchmark workflow reported successful detection and then skipped `Run: ${{ matrix.target }}`: https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/33917757666.

The exact page has been validated locally on both master and refreshed dependency stacks, but the PR cannot obtain a hosted benchmark artifact until this support-file detection change lands.

PR #1836, opened 2026-09-05
Report ESS/s and R-hat for the Stan backend in the FitzHugh-Nagumo benchmark

singhharsh1708:
The FitzHugh-Nagumo benchmark now reports bulk ESS/s and R-hat for the Stan backend, which previously reported only wall time.
`display_stan_timing` printed Stan's per-chain timing from the CSV but returned nothing, so there was no sampling time to divide the ESS by. It now returns the seconds on the `(Sampling)` line.
`summarystats` already computes `ess_bulk` and `rhat` for the Stan chain through the same MCMCDiagnosticTools path as the Turing chain, so both backends are on one estimator. Its `ess_per_sec` column is `missing` for Stan because that chain carries no start and stop time, which is why the rate is computed here rather than read off.
Verified locally against cmdstan 2.39.0 at 800 samples: Stan reports minimum ESS/s 75.6 over 3.61s of sampling, Turing 104.4 over 3.81s. Both benchmarks run a single chain, so R-hat here is the within-chain split diagnostic only.
This relies on `stan_inference` returning an `MCMCChains.Chains`, which holds for the pinned DiffEqBayes 3.13.0. On 3.16.1 it returns a DataFrame instead and this will need revisiting.

### DIFF
--- a/benchmarks/BayesianInference/DiffEqBayesFitzHughNagumo.jmd
+++ b/benchmarks/BayesianInference/DiffEqBayesFitzHughNagumo.jmd
@@ -14,21 +14,24 @@ using Plots, StaticArrays, Turing, LinearAlgebra
 ```
 
 ```julia
-"""Display ESS/s (effective samples per second) from a Turing chain."""
+"""Display bulk ESS/s and R-hat for a chain, given its sampling time."""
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
we want to bump this

PR #1878, opened 2026-09-13
DAE/fekete: strip investigation narrative from the document

ChrisRackauckas-Claude:
## Summary

Three commits to `benchmarks/DAE/fekete.jmd` on `fix/dae-runtime`:

- **Strip the investigation narrative** (1111 → 808 lines): removes dated reproduction logs, "used to follow… Removed" notes, per-block cost annotations, and the second MTK diagnostic chunk.
- **Make the MTK initialization consistent**: the algebraic multipliers `lam_i` carried a default `= 0.0`, which ModelingToolkit treats as a hard initial condition; `λ(0) = 0` contradicts the acceleration-level constraint (consistent value ≈ −4.75), so every MTK solve returned `InitialFailure`. Leaving `lam` unprescribed makes initialization consistent — verified: `Success`, residual 3.6e-15. Integration still goes `Unstable` near t ≈ 4 for every solver tried (FBDF, QNDF, NordsieckBDF at t ≈ 4.04; Rodas5P at t ≈ 4.31), so the MTK form stays excluded; the section text and diagnostic chunk now show only that remaining failure.
- **Drop the out-of-place Jacobian workaround**: the Manifest already pins OrdinaryDiffEqDifferentiation v3.9.0, whose `get_fresh_jacobian` branches on `isinplace` — verified `get_fresh_jacobian` on a RadauIIA5 integrator returns a 160×160 matrix under plain `AutoSpecialize`, and aborting solves return `Unstable` with diagnostics instead of throwing. The out-of-place method, its comment, and the `FullSpecialize` wrapper are removed; the block now matches master.

Merging into `fix/dae-runtime` updates #1670.

#### Test plan
- [x] MTK init solve returns `Success` with residual 3.6e-15 (Julia 1.11.9, project Manifest)
- [x] MTK solver sweep confirms the remaining `Unstable` failure near t ≈ 4
- [x] `get_fresh_jacobian` returns a matrix and aborting mass-matrix solves return retcodes, no wrapper exception
- [ ] Weave renders

🤖 Generated with Devin CLI (model: swe-2-max) — local session understood-abrosaurus

### DIFF
--- a/benchmarks/DAE/fekete.jmd
+++ b/benchmarks/DAE/fekete.jmd
@@ -36,11 +36,9 @@ We benchmark three formulations:
    directly to ModelingToolkit, which uses `structural_simplify` to
    automatically perform index reduction and generate an index-1 DAE.
    This benchmarks MTK's symbolic transformation pipeline on a large-scale
-   constrained mechanical system. **This formulation is built below but is
-   currently excluded from the work-precision diagrams** — as of 2026-08-23 it
-   fails to initialize and, when initialization is forced, goes unstable after
-   0.4% of the time span. See "MTK Index-Reduced Formulation: Currently
-   Dropped", which reproduces the failure.
+   constrained mechanical system. It is currently excluded from the
+   work-precision diagrams because it does not solve — see "Why the MTK
+   index-reduced formulation is excluded" below.
 
 Reference: Bendtsen, C., Thomsen, P.G.: Numerical solution of differential
 algebraic equations. IMM-DTU, Tech. Report (1999). Available at the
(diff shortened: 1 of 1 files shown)


benchmarks/DAE/fekete.jmd
```diff
@@ -698,35 +674,27 @@ plot(ref_sol, idxs = [121, 122, 123, 124, 125],
...
+The MTK problem built above is not benchmarked because it does not solve.
+Defaults on unknowns (`p1_1(t) = y0[1]`, `q1_1(t) = 0.0`, `lam1(t) = 0.0`)
+are treated by ModelingToolkit as hard initial conditions — 120 prescriptions
+for the 60 unknowns initialization is free to choose — and they are
+inconsistent: `λ(0) = 0` contradicts the acceleration-level constraint that
+determines λ (the reference solution has λ → −4.75). So initialization
+stalls and every solver returns `InitialFailure` at t = 0. Forcing
+initialization (`BrownFullBasicInit()`, or prescribing only the positions)
+gets past t = 0 but exposes a second, independent failure: the integration
+goes `Unstable` near t ≈ 4 of the [0, 1000] span for every solver tried.
```
ChrisRackauckas: Okay just fix it?

benchmarks/DAE/fekete.jmd
```diff
@@ -313,20 +311,9 @@ function fekete_jac!(J, y, p, t)
...
-# `f.jac(u, p, t)` regardless of whether the problem is in-place. Without this
-# method that diagnostic throws instead of printing, which turns an ordinary
-# failed work-precision point into a hard error that kills the whole weave.
-# (This is an upstream bug, fixed in OrdinaryDiffEqDifferentiation v3.9.0 --
-# `get_fresh_jacobian` there branches on `isinplace` and calls `calc_J!`. This
-# folder's Manifest pins v3.7.0, which does not. The workaround here is
-# version-independent, so it stays correct either way; verified in isolation on
-# 2026-08-24 against both v3.7.0 (throws without it) and v3.10.0.)
+# Out-of-place method: the instability diagnostic calls `f.jac(u, p, t)` when
+# a solve aborts, and on OrdinaryDiffEqDifferentiation < v3.9.0 it throws
```
ChrisRackauckas: just use a newer version?

ChrisRackauckas-Claude: Done in 9ee6e2b — the Manifest already pins OrdinaryDiffEqDifferentiation v3.9.0, so the out-of-place method, its comment, and the FullSpecialize wrapper are deleted; the block now matches master. Verified locally: get_fresh_jacobian on a RadauIIA5 integrator returns a 160×160 matrix under plain AutoSpecialize, and aborting solves return Unstable with diagnostics instead of throwing.

PR #1905, opened 2026-09-23
StiffDDE/Oregonator_wpd: tolerances for the corrected (oscillating) RADAR5 Oregonator

ChrisRackauckas-Claude:
> Please ignore until reviewed by @ChrisRackauckas.

**Blocked on https://github.com/SciML/DiffEqProblemLibrary.jl/pull/233 being merged and DDEProblemLibrary 0.1.10 being registered.** After that, run `Pkg.update("DDEProblemLibrary")` in `benchmarks/StiffDDE` and add the Manifest change to this PR. Until then CI would weave this page against 0.1.9, which still has the bug.

## Root cause

The published reference solution is flat (u1 ≈ 0, u2 ≈ 1e-5 on [0, 100.5]) because `DDEProblemLibrary.prob_dde_RADAR5_oregonator` has the wrong sign on `k₁ A u₂` in the u1 equation (it shared `a = -k₁*A*u[2] - k₂*u[1]*v` between both components). RADAR5's `dr-oregon.f` has `F(1) = +kM1*A*Y(2) - ...`. The page's own equation is correct. Fixed upstream in https://github.com/SciML/DiffEqProblemLibrary.jl/pull/233 (with a test that fails before and passes after).

With the correct RHS the solution spans about 1e-11 to 1e-5 (u2 decays from 1e-5, then periodic relaxation spikes). The page's absolute tolerances (1e-4 to 1e-11) sit far above that scale. In a first run with only the library fix, every solver in the high-tolerance sets gave the same final error of about 1.8e-7, which means none of them resolved the spikes.

## Fix (this PR)

- Scale absolute tolerances with the solution as the RADAR5 driver does (`ATOL = RTOL * 1e-9`): high tolerances `abstols = 1e-10:1e-13` (reltols unchanged, 1e-1:1e-4), low tolerances `abstols = 1e-14:1e-17` (reltols 1e-5:1e-8). The reference abstol goes from 1e-14 to 1e-21 (reltol 1e-14).
- Prose: the reference-solution paragraph explains the tolerance scaling, and the stiff vs non-stiff paragraph describes the new results. The old sentence said the explicit methods produce no points.

This changes the tolerances being benchmarked. That is the judgment call to review.

## Evidence

The page was run with the fixed DDEProblemLibrary dev'd into `benchmarks/StiffDDE`. The Manifest change is not committed.

```
~/.juliaup/bin/julia +1.11 --threads=auto --project=. benchmark.jl benchmarks/StiffDDE/Oregonator_wpd.jmd   # EXIT 0
```

- Reference plot (fig 2): u2 decays from 1e-5 to about 3e-7 by t ≈ 20, then spikes periodically about every 4 time units until t = 100.5. The published figure is flat.
- High-tolerance Rosenbrock (final error): errors now span 1e-10 to 1e-7 depending on tolerance. Before, every point was ~1e-16 (flat solution); with only the library fix, every point was ~1.8e-7.
- Stiff vs non-stiff: `Tsit5`, `BS3` and `DP5` now have points. Before, they appeared only in the legend. `DP5` at the loosest tolerance takes about 5 s.
- Low tolerances: errors span about 1e-16 to 1e-8.

Timings come from a heavily loaded shared machine (load average 300+ on 128 cores), so they are noisy.

## Not verified

The page has not been run against a registered DDEProblemLibrary 0.1.10, which does not exist yet.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (model: claude-opus-5-5[1m])

https://claude.ai/code/session_01Vrsu4PESdABpNaxYiTBVfd


pankgeorg (commented):
VERDICT: CHANGES

- **P1 — commit the corrected problem dependency before merging.** [`benchmarks/StiffDDE/Oregonator_wpd.jmd:48`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cfccf8430bcf0e4eeb16b4ceb133731db64167c6/benchmarks/StiffDDE/Oregonator_wpd.jmd#L48) still solves the imported library problem, while [`benchmarks/StiffDDE/Manifest.toml:252`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cfccf8430bcf0e4eeb16b4ceb133731db64167c6/benchmarks/StiffDDE/Manifest.toml#L252) pins `DDEProblemLibrary 0.1.9` (`dbb4929…`). The required RHS sign correction in [DiffEqProblemLibrary #233](https://github.com/SciML/DiffEqProblemLibrary.jl/pull/233) is still open. Consequently a clean checkout does not run the corrected oscillatory model described by the new prose. A successful run with a locally developed dependency does not validate the committed environment. Update the manifest to a released version containing the fix, then obtain a successful CI weave of this page.
- [`benchmarks/StiffDDE/Oregonator_wpd.jmd:60`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cfccf8430bcf0e4eeb16b4ceb133731db64167c6/benchmarks/StiffDDE/Oregonator_wpd.jmd#L60): the tolerance change itself is sensible: both sweeps now use `abstol = 1e-9 * reltol`, so absolute error control no longer dwarfs the small solution components. The tighter reference is appropriate once the corrected RHS is available.

Validation: inspected the committed manifest and upstream dependency PR, and checked the tolerance arithmetic. No benchmark or full suite run locally.

Reviewed head: `cfccf8430bcf0e4eeb16b4ceb133731db64167c6`.

CI comparison (checked 2026-09-24; each row is a failed PR job):

| Job | PR | Default branch |
|---|---|---|
| `Runic / Runic Format Check` | [FAILURE](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/35886760438/job/107268758185) | [SUCCESS](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/35977591365/job/107561632968) — latest run |

All these PR Runic logs report formatting in unchanged `cutest_benchmark_utils.jl`. That failure also occurred on the [September 23 base run](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/35855371838), but **the latest default-branch Runic run is green**; it should not be described as currently red on default.

Cancelled checks are incomplete validation, not passing or red test results:

| Job | PR | Latest matching default job |
|---|---|---|
| `Run: benchmarks/StiffDDE/Oregonator_wpd.jmd` | [CANCELLED](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/35886759334/job/107277944857) | [CANCELLED](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/35836030151/job/107099724184); absent from newest workflow run |

🤖 Posted for @pankgeorg by an AI agent — harness: Codex CLI 0.155.0 · model: gpt-6-astra
Session: local Codex session 01a0d371-cd25-7b23-9706-a1a1fa644f9f

PR #1909, opened 2026-09-23
DAE and NonStiffBVP: fix benchmark bugs, document solver failures, note upstream MIRK/DASKR bugs

ChrisRackauckas-Claude:
> Please ignore until reviewed by @ChrisRackauckas.

## Summary

Fixes and documents the suspicious results flagged in the DAE and NonStiffBVP benchmark pages: five benchmark-correctness fixes, four pages that now document genuine solver failures instead of showing unexplained missing points, and one upstream BoundaryValueDiffEq bug reduced to an MWE and fixed in SciML/BoundaryValueDiffEq.jl#648.

Validation entrypoint (CI weave), one page at a time:

```sh
JULIA_DEPOT_PATH=$HOME/sandbox/depot_fix5p TMPDIR=$HOME/tmp \
  ~/.juliaup/bin/julia +1.11 --threads=16 --project=. benchmark.jl benchmarks/<folder>/<page>.jmd
```

### `DAE/LinearDAE` — input switch at t = 0.1 now reinitializes λ

The mass-spring subsystem's input force jumps from 10 to 0.1·e⁻⁵ᵗ at t = 0.1, and the Lagrange multiplier λ depends algebraically on the input, so λ jumps too. Any adaptive step crossing t = 0.1 carries an O(jump) error in λ however short the step, so solvers could not get past the switch. The page now stops at t = 0.1 (`tstops`), switches the input via a `DiscreteCallback`, and resets λ to the value satisfying the algebraic equation under the new input. Weave log: exit 0.

### `DAE/slider_crank` — configuration-dependent mass matrix in all forms

The residual and mass-matrix forms froze the generalized mass matrix `AM(p)` at t = 0, capping achievable tolerance near 1e-7 while the MTK form used `AM(p)` at the current state. All three forms now evaluate `AM(p)` at the current state, and the velocity-level forcing term `γ = -(dGP/dt)·v` is evaluated analytically instead of by finite differences. Weave log: exit 0.

### `DAE/wheelset` — φ-dependent Euler-equation coupling in all forms

The mass-matrix and MTK forms froze the 3×3 Euler mass block (which depends on φ through cos φ, sin φ) at t = 0; the frozen reference was off from the IVP Test Set values at t = 10 by up to ~3%. The coupled block is now solved in closed form inside `f` at the current state, so the mass matrix is constant (`diag(I₅, MR·I₃, I₃, 0₆)`) and the formulation is exact. Weave log: exit 0.

### `DAE/andrews_mechanism` — MTK initialization and state selection; MTK form added to the WPD

Every MTK variable had a fixed initial value, making initialization overdetermined (warning suppressed). Now only β and β̇ get initial conditions. Index reduction had picked ε as the free angle, whose constraint-Jacobian block is singular at t = 0 (Θ = 0 makes the β and Θ columns parallel, cond ~1e16) so every solver failed at the first step; `state_priority` now selects β/β̇ (cond 3.4–5.0 along the trajectory). Rodas5P, Rodas4P, FBDF and NordsieckBDF now reach t = 0.03 matching the PSIDE positions to 2e-8..7e-6 and are included in the work-precision diagram against an MTK reference. Weave log: exit 0 (expected RADAU singular-matrix diagnostics appear in the output and are documented).

### `DAE/charge_pump` — exact solution as reference; DASKR failures reported

Charge conservation and Q_S = Q_D reduce the problem to a scalar monotone equation for U2 at each t (U3 = 4·U2, index-2 current I = -dY_T1/dt). The WPDs now measure errors against this exact solution instead of an IDA solve at 5e-4 whose current was off by up to 2.8e-3 A; the exact solution matches GAMD y1..y8 exactly and y9 to 6.1e-6. A retcode table documents that DASKR fails at reltol 1e-1 (HMIN at t = 1.01e-6) and stops at the first 500-step return at 1e-2/1e-3 under `save_everystep = false` — the DASKR.jl wrapper bug below. The L2 timeseries WPD now saves every step (before it compared only t0 and t_end). Weave log: exit 0 with the documented DASKR failures in the output.

### `DAE/fekete` — DASKR fails every tested pair; radau stops at t ≈ 1.18 on the tightest pair

A diagnostic sweep prints retcode/t_final for all tolerance pairs. DASKR: `MaxIters` with `t_final = 0` at the loose pairs (the 500-step wrapper bug below), `Failure`/`ConvergenceFailure` at the tight pairs — Fortran-level failures on this index-2 formulation, which DASKR is not designed for. `radau()` succeeds on all pairs except (abstol 1e-10, reltol 1e-7), where it stops at t ≈ 1.18 with `DtLessThanMin`; passing the index-2 variable counts (`DIMOFIND*VAR`) makes it stop earlier, at t ≈ 0.60. The generated page shows the printed table verbatim. Weave log: exit 0.

### `DAE/TransistorAmplifier` — RADAU "MATRIX IS REPEATEDLY SINGULAR" at X = 0.0126

At (abstol 1e-5, reltol 1e-1) both Fortran RADAU and RADAU5 exit at t = 0.0126 with a repeatedly singular matrix, while Julia `RadauIIA5()` completes the interval (304 steps). Tighter pairs complete. A diagnostic block prints all three retcodes; the prose explains which WPD points are missing and why. Weave log: exit 0.

### `DAE/ROBERDAE`, `DAE/ChemicalAkzoNobel`, `DAE/NANDGateProblem` — DASKR "500 STEPS TAKEN" early stops

DDASKR returns `IDID = -1` every 500 internal steps; the DASKR.jl wrapper in this manifest ends the solve there instead of resuming, so any `save_everystep = false` solve needing > 500 steps returns `MaxIters` with only the initial point saved and contributes no WPD point. Measured with the pages' own problems on the released wrapper:

- ROBERDAE: the three tightest low-tolerance pairs (abstol 1e-10..1e-12) → `MaxIters`, `t_end = 0`; looser pairs finish in < 500 steps.
- ChemicalAkzoNobel: the two tightest pairs → `MaxIters`.
- NANDGateProblem: every pair except the loosest → `MaxIters`, `t_end = 0`. Its `dae_ref_sol` uses `save_everystep = true` (intermediate-output mode) and is unaffected.

With SciML/DASKR.jl#122 applied, all of the above return `Success` to `t_end`. The pages now explain the missing points and reference that draft fix — no new DASKR issue was filed. These pages' code chunks are byte-identical to the version that produced the committed weave output; the changes are prose only.

### `NonStiffBVP/orbital` — MIRK2/MIRK4 anomaly is an upstream BoundaryValueDiffEq bug

Published: MIRK2 ≈ 210 s / 2.34 GiB, MIRK4 ≈ 26.5 s / 6.90 GiB, vs MIRK3 ≈ 3.1 s and MIRK5/6 < 1 s. Reduced to a BoundaryValueDiffEq-only MWE (`orbital-mwe.jl`), which reproduces it.

Root cause: after `mesh_selector!` redistributes the mesh, `__perform_mirk_iteration` seeds the next nonlinear solve by interpolating the previous solution at each new mesh point — but `interp_eval!` writes into `cache.y₀.u` in place while later interpolation points still read old-mesh values from it. On refinement the interpolated guess is corrupted and Newton restarts almost from scratch: MIRK4's refinement iteration began at ‖F(u₀)‖∞ ≈ 9.1e6 and took 1195 Newton iterations; with the fix it begins at ≈ 7.8e-3 and converges in 2.

Fixed upstream: **SciML/BoundaryValueDiffEq.jl#648** (draft) — evaluate interpolants into a fresh array, copy back after the loop; regression test fails 62/89 on master, passes 89/89 with the fix. Local orbital timings with the fix:

| solver | before | after |
|---|---|---|
| MIRK2 | ~483 s / 2.5 GiB | ~34 s / 516 MiB |
| MIRK3 | ~6.7 s / 347 MiB | ~1.4 s / 130 MiB |
| MIRK4 | ~46 s / 8.6 GiB | ~0.60 s / 83 MiB |
| MIRK5/6 | ~0.7–1.1 s | unchanged |

The page keeps the same solves and adds a section explaining that the timings include the upstream bug, with the PR link. MIRK2's remaining cost is genuine: its low order needs a ~2900-point mesh at this tolerance. Code chunks unchanged; prose-only edit.

## Not fixed / not verified

- **DASKR 500-step wrapper bug** — fixed upstream in SciML/DASKR.jl#122 (draft, opened in this work session); not merged, so the pages document the failures and reference the PR.
- **MIRK interpolation bug** — fixed in SciML/BoundaryValueDiffEq.jl#648 (draft); the orbital page still shows the buggy timings until that merges and releases.
- **fekete tight-pair DASKR failures** (`Failure`, `ConvergenceFailure`) — Fortran-level limitations on an index-2 problem; documented on the page, no upstream fix attempted.
- **radau `DtLessThanMin` on fekete** — documented; ODEInterface RADAU behavior, not pursued further.
- **Prose-only pages** (ROBERDAE, ChemicalAkzoNobel, NANDGateProblem, orbital) — code chunks verified byte-identical to the previously woven versions; not re-woven.
- MIRK test suite (`test/Core/mirk_basic_tests.jl`) run on the fix branch: in progress at PR-open time; the new discriminating testset passes 89/89 (vs 62 failures on master).

🤖 Generated with [Devin](https://devin.ai) (model: SWE-2 High)
Devin CLI 3000.11.1 — local session `1e449d1a-c91c-4c3d-bd02-6aae6f54552f` (no shareable URL; transcript in `~/.local/share/devin/cli/sessions.db`). Earlier commits on this branch were authored by Claude Code (session https://claude.ai/code/session_01Vrsu4PESdABpNaxYiTBVfd) and keep their original trailers.


pankgeorg (commented):
VERDICT: MERGE

The mathematical corrections are consistent with the equations they replace. No blocking finding in this diff.

- [`benchmarks/DAE/slider_crank.jmd:361`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cf4a7cbf7366d28d21ea54f459e0b34142e64431/benchmarks/DAE/slider_crank.jmd#L361) and [`benchmarks/DAE/slider_crank.jmd:384`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cf4a7cbf7366d28d21ea54f459e0b34142e64431/benchmarks/DAE/slider_crank.jmd#L384): the mass-matrix RHS now solves with `AM(q)` at the current configuration, while the residual uses that same current matrix multiplying the acceleration. The constant identity/zero mass matrix is therefore an algebraic reformulation rather than a frozen-mass approximation. The analytic initial constraint derivative has the expected quadratic-velocity and mixed terms.
- [`benchmarks/DAE/wheelset.jmd:560`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cf4a7cbf7366d28d21ea54f459e0b34142e64431/benchmarks/DAE/wheelset.jmd#L560): solving the triangular Euler block gives `θ̈=r₉/(LI2*cos(φ))`, `φ̈=r₁₀/LI2`, `β̈=r₁₁/LI1-sin(φ)*θ̈`, matching the residual coefficients and the updated symbolic equations.
- [`benchmarks/DAE/charge_pump.jmd:409`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cf4a7cbf7366d28d21ea54f459e0b34142e64431/benchmarks/DAE/charge_pump.jmd#L409): zero initial conserved charges and identical source/drain charge functions imply `U3=4U2`. The scalar root and implicit derivative then produce `I=-dQ_G/dt`. I checked conservation, finite-difference current derivatives away from switching points, and the endpoint values. The left-limit convention at pulse breakpoints is stated explicitly.
- [`benchmarks/DAE/LinearDAE.jmd:347`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cf4a7cbf7366d28d21ea54f459e0b34142e64431/benchmarks/DAE/LinearDAE.jmd#L347): the callback changes the forcing parameter and recomputes the algebraic multiplier while preserving differential states; stopping at the switch and saving both reference limits is consistent with a discontinuous multiplier. [`benchmarks/DAE/andrews_mechanism.jmd:297`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cf4a7cbf7366d28d21ea54f459e0b34142e64431/benchmarks/DAE/andrews_mechanism.jmd#L297) gives only the free coordinate and velocity fixed initial values, with guesses for dependent coordinates. The page correctly warns that its formulations' errors cover different variables; their error-axis values should not be interpreted as equal physical accuracy.
- [`benchmarks/NonStiffBVP/orbital.jmd:125`](https://github.com/SciML/SciMLBenchmarks.jl/blob/cf4a7cbf7366d28d21ea54f459e0b34142e64431/benchmarks/NonStiffBVP/orbital.jmd#L125): the aliasing explanation is supported by MIRK's `interp_eval!`/`sum_stages!`, which read `cache.y₀.u[i]` while the old loop overwrites that same storage. The linked fix stages the outputs before copying them back. I verified the mechanism, not the quoted speedups.

Validation: Julia 1.12.7, one memory-capped process at a time, using extracted mathematical functions only: **35/35** slider-crank/wheelset consistency checks and **33/33** charge-conservation/current/endpoint checks passed. No benchmark, full suite, or MTK integration run locally. The two added direct dependencies are used by the reference wrapper and symbolic callback; the manifest changes only its project hash. The remaining printed solver timings/failure thresholds are not independently reproduced by this review.

Reviewed head: `cf4a7cbf7366d28d21ea54f459e0b34142e64431`.

CI comparison (checked 2026-09-24; each row is a failed PR job):

| Job | PR | Default branch |
|---|---|---|
| `Runic / Runic Format Check` | [FAILURE](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/35927725789/job/107406711404) | [SUCCESS](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/35977591365/job/107561632968) — latest run |

All these PR Runic logs report formatting in unchanged `cutest_benchmark_utils.jl`. That failure also occurred on the [September 23 base run](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/35855371838), but **the latest default-branch Runic run is green**; it should not be described as currently red on default.

Still pending: [Run: benchmarks/DAE](https://github.com/SciML/SciMLBenchmarks.jl/actions/runs/35927725300/job/107406907217) (queued).

🤖 Posted for @pankgeorg by an AI agent — harness: Codex CLI 0.155.0 · model: gpt-6-astra
Session: local Codex session 01a0d371-cd25-7b23-9706-a1a1fa644f9f

